PR #3, opened 2020-07-10
WIP: Do not assume isoparametricity (iso-parametric-ness?)

inducer:
(IMO) the way that this should go is that the there should be a separate meshmode `Discretization` in `DGDiscretizationWithBoundaries` to capture geometry info. This discretization would then be used to compute geometry-related info, such as metric derivatives and area/surface elements. It should be straightforward to handle this by introducing a separate `quadrature_tag` (which this reveals to be mis-named, ugh).

The net benefit of this would be to enable order 0 DG (aka finite volume). It's needed because order 0 (one DOF per triangle, i.e. strictly isoparametric) doesn't do great at representing triangles. But we may like running subparametrically anyhow. For example, this permits a very easy way of "sectioning off" affinely-transformed element groups from ones that have non-affine mappings, where geometry calculations (and mass matrix inversions, even with WADG) are cheaper on affinely transformed elements.

NB: This sits on top array context work, which lives
- [here](https://gitlab.tiker.net/inducer/meshmode/-/merge_requests/80) in meshmode
- [here](https://gitlab.tiker.net/inducer/grudge/-/merge_requests/71) in grudge
so it should merge after those.

cc @lukeolson @mtcam @alexfikl

### DIFF
--- a/test/test_grudge.py
+++ b/test/test_grudge.py
@@ -251,7 +251,7 @@ def test_2d_gauss_theorem(ctx_factory):
     ])
 @pytest.mark.parametrize("op_type", ["strong", "weak"])
 @pytest.mark.parametrize("flux_type", ["central"])
-@pytest.mark.parametrize("order", [3, 4, 5])
+@pytest.mark.parametrize("order", [0, 3, 4, 5])
 # test: 'test_convergence_advec(cl._csc, "disk", [0.1, 0.05], "strong", "upwind", 3)'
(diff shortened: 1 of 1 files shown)


inducer:
FWIW, there's nothing here yet other than a failing test. :)

inducer:
@lukeolson invited you to the repo so you can push here.

thomasgibson:
Closing since https://github.com/inducer/grudge/pull/161 landed

PR #7, opened 2020-07-24
Fix wave-eager* example signs

majosm:
Sorry, I kept forgetting about this. 🙂 

I ran each example before and after the changes, and the results seem to be as expected (i.e., still stable, `u` unchanged, `v` different by a minus sign).


majosm:
@inducer Should be good to go (doesn't look like I can request a review here).

inducer:
> (doesn't look like I can request a review here).

Really? Weird. Wish you could. Maybe we can look into that.

inducer:
Thanks for working on this!

PR #12, opened 2020-08-17
Fix `map_if` handling of scalar conditions

alexfikl:
Scalar case wasn't handled correctly because `np.bool` is not an `np.number`.


inducer:
Thanks!

PR #13, opened 2020-08-17
Fix max norm handling

alexfikl:
Seemed to think `pymbolic.Max` worked on pairs instead of just one iterator.

### DIFF
--- a/test/test_grudge.py
+++ b/test/test_grudge.py
@@ -645,6 +645,46 @@ def test_function_symbol_array(ctx_factory, array_type):
     assert isinstance(norm, float)
 
 
+@pytest.mark.parametrize("p", [2, np.inf])
+def test_norm_obj_array(ctx_factory, p):
+    """Test :func:`grudge.symbolic.operators.norm` for object arrays."""
+
+    ctx = ctx_factory()
+    queue = cl.CommandQueue(ctx)
+    actx = PyOpenCLArrayContext(queue)
+
+    from meshmode.mesh.generation import generate_regular_rect_mesh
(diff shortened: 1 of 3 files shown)


test/test_grudge.py
```diff
@@ -645,6 +645,34 @@ def test_function_symbol_array(ctx_factory, array_type):
...
+            a=(-0.5,)*dim, b=(0.5,)*dim,
+            n=(8,)*dim, order=1)
+    discr = DGDiscretizationWithBoundaries(actx, mesh, order=4)
+
+    w = make_obj_array([1.0, 2.0, 3.0])[:dim]
+
+    sym_w = sym.var("w")
+    norm = bind(discr, sym.norm(p, sym_w))(actx, w=w[0])
+    logger.info("norm: %.5e", norm)
+    # assert abs(norm / area - 1.0) < 1.0e-2, norm / area
```
alexfikl, line 668: For some reason, the errors for the 2-norm were huge here, although it's just integrating a constant. Why is that?

inducer: Fixed and reenabled in 51e3b51c8813248ad5d9e362917ddba3da6bf6f8. Taking the absolute value is invalid... they're just a bunch of weighted sums of inner products of Lagrange polynomials, and there is no reason we should be allowed to interpret them nodally.

Nice catch!

inducer:
Could you take care of the conflicts?

alexfikl:
> Could you take care of the conflicts?

Done!

inducer:
Thanks for working on this!

PR #17, opened 2020-09-08
Py3 Upgrades

alexfikl:
Originally at https://gitlab.tiker.net/inducer/grudge/-/merge_requests/85

Changes made by running
```
find . -name "*.py" -exec pyupgrade --py36-plus {} +
```
and fixing some of the resulting flake8 errors.


inducer:
I pushed the missing commits from Gitlab, but it doesn't look like it updated the diff. Not sure what to do, maybe recreate the PR?

alexfikl:
> I pushed the missing commits from Gitlab, but it doesn't look like it updated the diff. Not sure what to do, maybe recreate the PR?

Opened a new one at #18 and it seems to be properly updated!

PR #25, opened 2020-10-26
Doc unit normal

MTCam:
Add some simple doc for unit normal getter.

### DIFF
--- a/grudge/eager.py
+++ b/grudge/eager.py
@@ -251,6 +251,11 @@ class EagerDGDiscretization(DGDiscretizationWithBoundaries):
 
     @memoize_method
     def normal(self, dd):
+        """Get unit normal to specified surface discretization, *dd*.
+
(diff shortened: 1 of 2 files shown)


grudge/eager.py
```diff
@@ -251,6 +251,11 @@ def weak_div(self, *args):
 
     @memoize_method
     def normal(self, dd):
+        """Get unit normal to specified surface discretization, *dd*.
+
+        :arg dd: a :class:`~grudge.sym.DOFDesc` as the surface discretization.
+        :returns: surface normal as :class:`~pymbolic.geometric_algebra.MultiVector`.
```
inducer, line 257: This doesn't seem right. While it's true that grudge relies on geometric algebra/exterior calculus internally, I'm not crazy enough to believe that I can get away with inflicting that on a broader user base. I just tried calling `discr.normal`, and I got an object array of `DOFArray`s back.

inducer:
Thx!

PR #30, opened 2020-11-16
Fix object array vectorization in interior_trace_pair

majosm:
Fixes scalar case (previously missing "else" condition").

Tested in illinois-ceesd/mirgecom#128


inducer (approved):

inducer:
Thanks!

PR #31, opened 2020-11-18
Install mpi4py for Github CI

inducer:


inducer:
Unnecessary, already being installed in https://github.com/inducer/grudge/blob/fb82104ef80c87e46233e15b03009f6454c87c72/requirements.txt#L2

PR #35, opened 2020-12-03
CI: Run using modern modepy

alexfikl:
Running with inducer/modepy#16 and inducer/meshmode#70.


test/test_grudge.py
```diff
@@ -243,7 +242,7 @@ def test_mass_surface_area(actx_factory, name):
 
         h_max = bind(discr, sym.h_max_from_volume(
             discr.ambient_dim, dim=discr.dim, dd=dd))(actx)
-        eoc.add_data_point(h_max, area_error)
+        eoc.add_data_point(h_max, area_error + 1.0e-16)
```
alexfikl, line 245: So this doesn't seem to be needed on master (see #36), so something's fishy. Maybe it's just a local issue (just switched everything to Python 3.9 and possibly other updates)

alexfikl: Looks like it's just me :cry:

PR #37, opened 2020-12-18
Add mass operator to EagerDGDiscretization

majosm:
Adds a `mass()` operator method to `EagerDGDiscretization` (and fixes a couple of little bugs).

### DIFF
--- a/grudge/eager.py
+++ b/grudge/eager.py
@@ -58,6 +58,7 @@ class EagerDGDiscretization(DGDiscretizationWithBoundaries):
     .. automethod:: weak_div
 
     .. automethod:: normal
+    .. automethod:: mass
     .. automethod:: inverse_mass
     .. automethod:: face_mass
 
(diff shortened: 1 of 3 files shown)


inducer:
Is this draft still? (if so, please mark as such) I ask because:

- No tests
- Does not seem to add a mass operator

majosm:
Debugging some test failures... might have fixed them now.

Testing via illinois-ceesd/mirgecom#180 at the moment. Not sure how to add tests for it here (I didn't see any for the other eager stuff that I could base off of).

grudge/symbolic/operators.py
```diff
@@ -486,7 +486,7 @@ def __init__(self, dd_in=None, dd_out=None):
         if dd_in is None:
             dd_in = prim.DD_VOLUME
         if dd_out is None:
-            dd_out = dd_in
+            dd_out = prim.DD_VOLUME
```
majosm, line 489: @inducer This change broke a few things. I fixed the ones that were causing test failures, but I'm not positive that I got everything. Is there anything else I should check to be more confident that this is ok?

inducer: IMO: Tests are what provides confidence. If it wasn't tested, you may assume it was broken and don't need to worry about it.

grudge/eager.py
```diff
@@ -83,6 +84,9 @@ def project(self, src, tgt, vec):
         :arg tgt: a :class:`~grudge.sym.DOFDesc`, or a value convertible to one
         :arg vec: a :class:`~meshmode.dof_array.DOFArray`
         """
+        if src == tgt:
```
inducer, line 87:
```suggestion
        src = sym.as_dofdesc(src)
        tgt = sym.as_dofdesc(tgt)
        if src == tgt:
```

inducer (commented):
LGTM. One minor suggestion, which I'll apply.

PR #41, opened 2021-01-27
Handle projection of constants

majosm:
Tested in illinois-ceesd/mirgecom#180.

Seems to work, though it's maybe a little bit crude. It bypasses some of the from/to compatibility checks in `connection_from_dds`; I could implement this inside `connection_from_dds` instead, but I think then I would need to define a new connection type to return?

### DIFF
--- a/grudge/eager.py
+++ b/grudge/eager.py
@@ -33,6 +33,7 @@ from meshmode.dof_array import freeze, flatten, unflatten
 from grudge.discretization import DGDiscretizationWithBoundaries
 from grudge.symbolic.primitives import TracePair
 
+from numbers import Number
 
 __doc__ = """
 .. autoclass:: EagerDGDiscretization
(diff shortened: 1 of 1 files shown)


majosm:
Weird, CI is running out of memory...

inducer:
The updated PyOpenCL got built in conda forge today, and I merged https://github.com/inducer/loopy/pull/195.

Rerunning master: https://github.com/inducer/grudge/actions/runs/506776271

inducer:
The OOM message is a red herring. What it's trying to say is that pocl and openmpi disagree about the hwloc versions against which they were built. My guess for the root of the problem is that the CI images got updated to Ubuntu 20.x. See #42.

inducer:
Merging in master should get this going.

inducer (approved):

inducer:
Thanks!

PR #49, opened 2021-02-18
Make reference diff use elwise_linear loopy kernel

inducer:
@nchristensen This gets rid of the diff kernel as it currently exists. This is needed in order to get lazy eval to go. Are you OK with this going into master? (I ask because you've been working with the diff kernel as it currently exists. The point is that this current version essentially does one limited form of fusion "prematurely", when creating this form is likely better left to the transformed side of things. I imagine that essentially the same kernel would reemerge post-fusion on the "other side".)

cc @matthiasdiener @kaushikcfd

### DIFF
--- a/grudge/execution.py
+++ b/grudge/execution.py
@@ -25,7 +25,6 @@ from numbers import Number
 import numpy as np
 
 from pytools import memoize_in
-from pytools.obj_array import make_obj_array
 
 import loopy as lp
 import pyopencl as cl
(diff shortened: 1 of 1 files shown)


nchristensen:
I don't think this interferes with anything on my end. My plan is to route the elwise_linear calls through diff_batch_assign. Would this cause any problems with lazy eval? https://github.com/nchristensen/grudge/blob/c0c295a7ca43efeb7c1a7cda3cff76d125545efd/grudge/execution.py#L269

inducer:
> Would this cause any problems with lazy eval?

Well, this PR effectively kills off the batched differentiation kernel in favor of the elementwise matvecs. According to the code you linked, this would very much interfere with what you're doing, because there's no longer a batched diff kernel to dispatch to.

This is done because its multiple-vectors-at-a-time operating principle of batched diff isn't particularly great for lazy eval. It *could* be made to work, but as I wrote, I think it's easier if we keep the input to the lazy eval DAG simple and then reintroduce this form of fusion later.

I will still merge this though, since getting lazy eval going is fairly important. If nothing else, you could stay back on a branch where batched diff is a thing.

PR #52, opened 2021-03-07
Do not use CL-specific max, min, sum

kaushikcfd:


grudge/execution.py
```diff
@@ -150,25 +149,22 @@ def map_elementwise_max(self, op, field_expr):
...
 
     def map_nodal_sum(self, op, field_expr):
-        # FIXME: Could allow array scalars
-        # FIXME: Fix CL-specific-ness
+        actx = self.array_context
         return sum([
-                cl.array.sum(grp_ary).get()[()]
-                for grp_ary in self.rec(field_expr)
-                ])
+            actx.to_numpy(actx.np.sum(grp_ary))[()]
```
inducer, line 154: The `to_numpy` worries me in terms of lazy eval. Can we avoid that?

kaushikcfd: Yep, agreed. Done here: 98e6acf

grudge/execution.py
```diff
@@ -150,26 +149,19 @@ def map_elementwise_max(self, op, field_expr):
...
     def map_nodal_sum(self, op, field_expr):
-        # FIXME: Could allow array scalars
-        # FIXME: Fix CL-specific-ness
-        return sum([
-                cl.array.sum(grp_ary).get()[()]
-                for grp_ary in self.rec(field_expr)
-                ])
+        actx = self.array_context
+        return actx.np.sum(actx.np.stack([actx.np.sum(grp_ary)
+                                          for grp_ary in self.rec(field_expr)]))
```
inducer, line 154: The `stack` here goes in the wrong direction IMO: It'll expend lots of memory bandwidth producing a temporary array that'll just get thrown away. I think the outer reductions are fine here. For `min` and `max`, you probably want two nested calls to `actx.np.{min,max}`: One two reduce the arrays, and another across arrays.

inducer:

thomasgibson:
FWIW: I beleive "New world Grudge" #74 solves this: https://github.com/inducer/grudge/blob/thg/remove-symbolics/grudge/op.py#L1018-L1049

kaushikcfd:
@thomasgibson: thanks for taking a look here. I think the `nodal_max` in the linked code still contain calls to `np.max` which would fail in the lazy evaluation mode.

grudge/execution.py
thomasgibson: How do you ensure the outer reduction actually knows what to do with the inner reductions (that's what I see the `actx.np.stack` doing here)? Specifically the problem across a group of arrays. AFAICT, lists and generators are something `actx.np.{sum,min,max}` do not like.

thomasgibson:
> @thomasgibson: thanks for taking a look here. I think the `nodal_max` in the linked code still contain calls to `np.max` which would fail in the lazy evaluation mode.

@kaushikcfd Thanks for pointing that out, that slipped past me. Now I'm not sure how to avoid the use of `actx.np.stack`, because the outermost reduction (`actx.np.{sum,min,max}`) expects a particular type of object. Generators and lists are not compatible.

kaushikcfd:
stack and then reduce works, no?

cla.stack accepts lists in stack. pytato also supports list of array to be stacked.

```python
In [5]: a = cla.arange(queue, 10, 20, dtype=float)

In [6]: b = cla.arange(queue, 30, 40, dtype=float)

In [7]: cla.stack([a, b])
Out[7]: 
cl.Array([[10., 11., 12., 13., 14., 15., 16., 17., 18., 19.],
       [30., 31., 32., 33., 34., 35., 36., 37., 38., 39.]])
```

grudge/execution.py
kaushikcfd: Replied as a comment, but I don't see a problem why actx.np.stack shouldn't take a list of arrays.

thomasgibson: @kaushikcfd I think I misrepresented what I was trying to say. What I mean to say is this: I agree with your approach and I don't see another way to do it. The issue I'm referencing is completely removing `actx.np.stack` (which I had thought it was @inducer is suggesting, but maybe I'm wrong?). For example:
```python
actx.np.{min,max}([actx.np.{min,max}(grp_ary) for grp_ary in self.rec(field_expr)])
```
does _not _ work for the reason I described before.

thomasgibson:
> stack and then reduce works, no?
> 
> cla.stack accepts lists in stack. pytato also supports list of array to be stacked.
> 
> ```python
> In [5]: a = cla.arange(queue, 10, 20, dtype=float)
> 
> In [6]: b = cla.arange(queue, 30, 40, dtype=float)
> 
> In [7]: cla.stack([a, b])
> Out[7]: 
> cl.Array([[10., 11., 12., 13., 14., 15., 16., 17., 18., 19.],
>        [30., 31., 32., 33., 34., 35., 36., 37., 38., 39.]])
> ```

@kaushikcfd I replied above, but I was not talking about problems with passing a list to `actx.np.stack`. I was referencing problems with passing a list to `actx.np{min,max}`.

grudge/execution.py
kaushikcfd: Oops thanks @thomasgibson for clarifying it out. You know what, now that you pointed it out I think one way we could do is:
```python
from functools import reduce
return reduce(lambda acc, xi: actx.np.max(acc, actx.np.max(xi)), self.rec(field_expr), -np.inf) 
```

kaushikcfd: @thomasgibson: Is this logic being handled in your branch, or should we handle it here.

thomasgibson: I'll give it a try, but an initial attempt produced the following stacktrace:
```
a = -inf, b = 1.1942196769792337, out = None, queue = <pyopencl._cl.CommandQueue object at 0x7f8c7c36f7c0>

    def maximum(a, b, out=None, queue=None):
        """Return the elementwise maximum of *a* and *b*."""
    
        # silly, but functional
>       return if_positive(a.mul_add(1, b, -1, queue=queue), a, b,
                queue=queue, out=out)
E       AttributeError: 'float' object has no attribute 'mul_add'

../../pyopencl/pyopencl/array.py:2724: AttributeError
```

I'll try to see why. Basically the functions I wrote should more or less be identical to those that you've updated.

kaushikcfd: That's pyopencl being buggy: https://github.com/inducer/pyopencl/pull/457

kaushikcfd: FWIW, I felt this was the correct way to solve and added it here: dba33e3. Handling pyopencl comments now, to close this.

thomasgibson: I quick like this solution. Sounds like https://github.com/inducer/pyopencl/pull/457 needs to get merged :)

kaushikcfd: That one's again on me.. :laughing:

thomasgibson: np --- It's good to know there exists a clean solution to this problem. nice job!

inducer:
LGTM, thanks! I'll add that #74 (to land soon, hopefully) already does this right from the get-go (just checked).

PR #58, opened 2021-03-16
Remove usage of deprecated n parameter in generate_*_rect_mesh

majosm:
Depends on inducer/meshmode#133.

- [x] :warning: Point meshmode branch back to main in requirements.txt before merging.


inducer:
Thanks! (I had totally missed this.)

PR #59, opened 2021-03-20
Fix possible nan eoc in test_mass_surface_area

alexfikl:
Fix some fallout from inducer/meshmode#135. 

Not sure why it didn't fail already in that PR, but it's failing now in inducer/meshmode#136 :(


inducer:
Thx!

PR #62, opened 2021-04-06
Remove the use of deprecated functions/elements

thomasgibson:
This PR seeks to reduce the number of `DeprecationWarning`s by updating function calls and various other elements.


grudge/symbolic/operators.py
```diff
@@ -651,20 +654,32 @@ def matrix(self, afgrp, volgrp, dtype):
                     afgrp.nunit_dofs),
                 dtype=dtype)
 
-        from modepy.tools import UNIT_VERTICES
         import modepy as mp
+
+        vol_basis = volgrp.basis_obj()
+        faces = mp.faces_for_shape(volgrp.shape)
+
+        # NOTE: Assumes each face has the same quadrature/basis
```
inducer, line 662: Why not move this computation inside the face loop, below? You can zip that with the `face_vertex_indices` loop, or better yet, get the face vertex indices via the shape. `face_vertex_indices` in meshmode should probably be deprecated in favor of using more modepy (feel free to do here or file an issue).

grudge/symbolic/operators.py
```diff
@@ -651,20 +654,32 @@ def matrix(self, afgrp, volgrp, dtype):
...
+        face_quadrature = mp.quadrature_for_space(
+            # FIXME: How do we want to handle this in general?
+            # Do we always want a quadrature rule that exactly
+            # integrates 2N polynomials on the faces?
+            # For tensor-product and simplices?
+            mp.space_for_shape(faces[-1], 2*afgrp.order),
+            faces[-1]
+        )
```
inducer, line 670: Should assert `exact_to` condition. `order` doesn't always match that.

grudge/symbolic/operators.py
```diff
@@ -651,20 +654,32 @@ def matrix(self, afgrp, volgrp, dtype):
...
+
+        vol_basis = volgrp.basis_obj()
+        faces = mp.faces_for_shape(volgrp.shape)
+
+        # NOTE: Assumes each face has the same quadrature/basis
+        face_quadrature = mp.quadrature_for_space(
+            # FIXME: How do we want to handle this in general?
+            # Do we always want a quadrature rule that exactly
+            # integrates 2N polynomials on the faces?
+            # For tensor-product and simplices?
```
inducer, line 667: Why not? What downside do you see?

inducer (commented):
Thanks! Some comments on the face mass matrix below.

inducer:
Not sure why this appears to impacts diffusion accuracy in mirgecom. Maybe compare the matrices?

inducer:

grudge/symbolic/operators.py
thomasgibson: Wasn't sure if it would be potentially expensive to keep recomputing this for each face, but I guess it's okay?

Also BTW we don't need `face_vertex_indices` at all since each `face` knows how to associate the unit nodes (on the volume element) and itself

grudge/symbolic/operators.py
thomasgibson: Done

grudge/symbolic/operators.py
thomasgibson: I was thinking about over integration, but I think my latest commit addresses this issue.

grudge/symbolic/operators.py
```diff
@@ -651,20 +654,31 @@ def matrix(self, afgrp, volgrp, dtype):
...
+            assert face_quadrature.exact_to >= afgrp.order
+            face_basis = mp.basis_for_space(
+                mp.space_for_shape(face, afgrp.order),
+                face
+            )
+            matrix[:, iface, :] = mp.nodal_mass_matrix_for_face(
+                faces[iface], face_quadrature,
+                face_basis.functions, vol_basis.functions,
+                volgrp.unit_nodes, afgrp.unit_nodes,
+            )
```
thomasgibson, line 681: @inducer Could you confirm if you think this is the right decision here?

thomasgibson:
Thoughts on the PR?

grudge/symbolic/operators.py
```diff
@@ -651,20 +658,31 @@ def matrix(self, afgrp, volgrp, dtype):
...
+            # NOTE: Quadrature degree determined such that
+            # we either guarantee exact integration of degree 2N
+            # polynomials, or use a higher-order quadrature rule
+            # provided by the user (for example, when doing overintegration)
+            face_quad_degree = max(2*volgrp.order, afgrp.order)
+            face_quadrature = mp.quadrature_for_space(
+                mp.space_for_shape(face, face_quad_degree),
+                face
+            )
+            assert face_quadrature.exact_to >= afgrp.order
```
inducer, line 676:
```suggestion
            assert face_quadrature.exact_to >= face_quad_degree
```

grudge/symbolic/operators.py
```diff
@@ -651,20 +658,31 @@ def matrix(self, afgrp, volgrp, dtype):
...
+
+        vol_basis = volgrp.basis_obj()
+        faces = mp.faces_for_shape(volgrp.shape)
+
+        for iface, face in enumerate(faces):
+            # NOTE: Quadrature degree determined such that
+            # we either guarantee exact integration of degree 2N
+            # polynomials, or use a higher-order quadrature rule
+            # provided by the user (for example, when doing overintegration)
+            face_quad_degree = max(2*volgrp.order, afgrp.order)
```
inducer, line 671:
```suggestion
            face_quad_degree = 2*max(volgrp.order, afgrp.order)
```

grudge/symbolic/operators.py
```diff
@@ -651,20 +658,31 @@ def matrix(self, afgrp, volgrp, dtype):
...
-        # 1/0
+
+        vol_basis = volgrp.basis_obj()
+        faces = mp.faces_for_shape(volgrp.shape)
+
+        for iface, face in enumerate(faces):
+            # NOTE: Quadrature degree determined such that
+            # we either guarantee exact integration of degree 2N
+            # polynomials, or use a higher-order quadrature rule
+            # provided by the user (for example, when doing overintegration)
```
inducer, line 670: I don't understand the motivation here. Why is the volume being treated differently than the surface?

grudge/symbolic/operators.py
```diff
@@ -651,20 +658,31 @@ def matrix(self, afgrp, volgrp, dtype):
...
+            face_quad_degree = max(2*volgrp.order, afgrp.order)
+            face_quadrature = mp.quadrature_for_space(
+                mp.space_for_shape(face, face_quad_degree),
+                face
+            )
+            assert face_quadrature.exact_to >= afgrp.order
+            face_basis = mp.basis_for_space(
+                mp.space_for_shape(face, afgrp.order),
+                face
+            )
```
inducer, line 680: Why not ask the `afgrp` for its basis?

inducer (commented):
Thanks! A few questions/comments regarding the face mass matrix computation, the rest looks fine to me.

grudge/symbolic/operators.py
inducer: Ah. Still though, if it *does* have one, IMO we should use it. `if isinstance(afgrp, ElementGroupWithBasis)`...

grudge/symbolic/operators.py
thomasgibson: @inducer When you overintegrate, the `afgrp` is coming from the higher-order quadrature grid. This is just a note for why I computed `face_quad_degree` below. If I just took the order from the volume group (i.e. `face_quad_degree = 2*volgrp.order`, this causes the diffusion test in mirgecom to fail (because it's actually using overintegration.

Take a look at this test here to see the example I'm trying to handle `https://github.com/illinois-ceesd/mirgecom/blob/main/test/test_diffusion.py#L271`

If you think this can be handled in a different way, I'm open to suggestions.

grudge/symbolic/operators.py
thomasgibson: Okay! Will add this in

grudge/symbolic/operators.py
inducer: Thanks for poking at this. It turns out that it was much more broken than I thought, cf. https://github.com/inducer/modepy/pull/28. There should genuinely be two cases here, nodal and overintegrated.

inducer: (So, to be clear, I'm proposing that we base the work here on the new functionality in https://github.com/inducer/modepy/pull/28, so we'll need a modepy version lower bound here.)

grudge/symbolic/operators.py
```diff
@@ -651,20 +658,37 @@ def matrix(self, afgrp, volgrp, dtype):
...
+                face_basis = mp.basis_for_space(
+                    mp.space_for_shape(face, afgrp.order),
+                    face
+                )
+
+            matrix[:, iface, :] = mp.nodal_mass_matrix_for_face(
+                face, face_quadrature,
+                face_basis.functions, vol_basis.functions,
+                volgrp.unit_nodes, afgrp.unit_nodes,
+            )
```
thomasgibson, line 691: @inducer In light of https://github.com/inducer/modepy/pull/28, does this mean we should _always_ be using `nodal_quad_mass_matrix_for_face`? Or only when we have a quadrature-based element group on the faces?

thomasgibson: @inducer This doesn't work out of the box. The shape of the guy being returned from `nodal_quad_mass_matrix_for_face` just flat out does not match with that this method is expecting:
```python
matrix = np.empty(
        (volgrp.nunit_dofs,
            nfaces,
            afgrp.nunit_dofs),
        dtype=dtype)
```
And when I pass `afgrp._quadrature_rule()` through the method, I'm getting some einsum value error...
`ValueError: einstein sum subscripts string contains too many subscripts for operand 1`.

I will push my latest commit so you can try and see for yourself (https://github.com/inducer/grudge/pull/62/commits/5c047f6069fa6fb29be097754d04a454e711f564).

Run this to see exactly what I'm seeing:
```
py.test test_grudge.py -k 'test_improvement_quadrature' -x 
```

I will try to spend more time this afternoon digging into what's going on... but I'd appreciate it if you could try it and/or let me know if you see something immediately wrong.

inducer: > only when we have a quadrature-based element group on the faces?

Only then.

inducer: If you think about it, they are two different numerical procedures. In the overintegrated case, the quadrature is "exposed". In the nodal case, we use quadrature to compute integrals, but the input data for the matrix is in terms of Lagrange coefficients. For example, the latter needs to know what the trial basis is, the former does not.

inducer: > but I'd appreciate it if you could try it and/or let me know if you see something immediately wrong.

Needs https://github.com/inducer/meshmode/pull/156

grudge/symbolic/operators.py
```diff
@@ -651,20 +658,41 @@ def matrix(self, afgrp, volgrp, dtype):
...
+
+        n = volgrp.order
+        m = afgrp.order
+        vol_basis = volgrp.basis_obj()
+        faces = mp.faces_for_shape(volgrp.shape)
+        face_quad_degree = 2*max(n, m)
+
+        for iface, face in enumerate(faces):
+            try:
+                face_quadrature = afgrp._quadrature_rule()
```
inducer, line 672: It's not amazing that we have to reach into an underscored property here.

thomasgibson: So the simplest solution I can see here is to just rename `_quadrature_rule()` to `quadrature_rule()`. That seems fine to me.

On a related note: any comment on whether you think this is appropriate? The rationale for this try-except clause is that if a `afgrp` has a quadrature rule already associated with it, then we should just use that rather than create a new quadrature object. Thoughts?

alexfikl: The `unit_nodes` and `weights` are public methods on the element groups, so you could also just yank those and stick them into a `mp.Quadrature` directly.

Not sure if that's less hacky or more hacky though..

thomasgibson: I see what you mean. But it feels wasteful to keep instantiating new `mp.Quadrature` objects when there is one already available.

My second paragraph is more or less just stating that I'm thinking we should try to use the quadrature rule defining the `afgrp` (if there is one) rather than just keep creating new quadrature objects.

alexfikl: That's a fair point. I don't feel too bad about it because the results from this get memoized, so it's not a big cost.

grudge/symbolic/operators.py
```diff
@@ -651,20 +658,41 @@ def matrix(self, afgrp, volgrp, dtype):
...
+            except AttributeError:
+                face_quadrature = mp.quadrature_for_space(
+                    mp.space_for_shape(face, face_quad_degree),
```
alexfikl, line 675: Doesn't this lead to different quadrature orders?
* In the case where there is a `_quadrature_rule`, the order is `afgrp.order`
* In the case there isn't, the order is `2 * max(volgrp.order, afgrp.order)`

thomasgibson: In the case of this example in mirgecom: https://github.com/illinois-ceesd/mirgecom/blob/2a39aa880ff883774f9646e23746c682c151fd80/test/test_diffusion.py#L271
The over integration order on the `QuadratureSimplexElementGroup` I think is already taking into account what quadrature degree you want. (3*order). And so that's why I think we should use the underlying quadrature rule instead of creating one.

My understanding of these quadrature element groups is that the order is an explicit statement on the order of quadrature you want to use. Is this a correct statement?

Now the `except` clause is more or less handling the "non over integrated" case, with possibly different quadrature rules in the volume and face. And the calculation for the order ensures that we integrate the product of polynomials restricted to the face exactly, which is more or less what the "old" face mass routine in modepy does: https://github.com/inducer/modepy/blob/7ebb43bd7612adf55991c5535811d0eaf8f4368e/modepy/matrices.py#L358

So this is just a version of that which considers the possibility that face and volume quadrature may have different orders.

After thinking about this as I'm writing, I'm wondering if a more readable way is to do something along the lines of: `if isinstance(afgrp, QuadratureSimplexElementGroup):...` rather than try-except. I think that's technically equivalent to the try-except code here, but maybe it's more explicit?

grudge/symbolic/operators.py
thomasgibson: OK as long as you all don't think it's too much of an issue, I can be convinced it's okay :smile:

grudge/symbolic/operators.py
inducer: > My understanding of these quadrature element groups is that the order is an explicit statement on the order of quadrature you want to use. Is this a correct statement?

No. `order` tends to serve as a loose identifier of different rules within a family of quadrature rules. It certainly correlates with exactness degree, and sometimes that's exactly what you get, but in general you have to look at the `exact_to` attribute to get a promise of exactness.

My sense is that we should just recast the `NodalElementGroupBase` to make a modepy quadrature rule available (and, consequently, deprecate the `weights` attribute).

> But it feels wasteful to keep instantiating new mp.Quadrature objects when there is one already available.

Instantiating objects is cheap. Recomputing weights is not. (But see above.)

thomasgibson: > No. `order` tends to serve as a loose identifier of different rules within a family of quadrature rules. It certainly correlates with exactness degree, and sometimes that's exactly what you get, but in general you have to look at the `exact_to` attribute to get a promise of exactness.


OK. Thanks for the clarification.


> My sense is that we should just recast the `NodalElementGroupBase` to make a modepy quadrature rule available (and, consequently, deprecate the `weights` attribute).


OK. So basically make `_quadrature_rule` public?


> Instantiating objects is cheap. Recomputing weights is not. (But see above.)


Good to know.

thomasgibson: Uhg. Not all `modepy.Quadrature` objects have an `exact_to` method...

inducer: > OK. So basically make `_quadrature_rule` public?

Well, yes, but also move it up to `NodalElementGroupBase`.

inducer: > Edit: Oh. I'm getting smacked by the zero-dimensional quadrature rule again.

@thomasgibson asked in chat what the accuracy should be. I wasn't sure. `np.inf`? `None`?

Also, it might be worthwhile to centralize/standardize that 0D quadrature rule in modepy, so that we don't create multiple inconsistent versions of it.

alexfikl: > `np.inf`? `None`?

I kind of like `np.inf`. It makes sense, comparisons with some finite `exact_to` are valid and there's already the special case of `AttributeError` to say we don't know.

thomasgibson: What about a general quadrature rule constructed from a set of weights/nodes? To me, this seems almost impossibly difficult to determine. Which is why I get why there is no generic `exact_to` method for `modepy.Quadrature`. Perhaps what should happen is that `modepy.Quadrature` should not be allowed to be instantiated on its own (i.e. make it an abstract class); rather have only concrete subclass of it, where we actually know about the quad rule and where it comes from.

alexfikl: Yes, agreed on all of that. 

I was thinking more about adding a separate `PointQuadrature` for 0d things, which can have an `np.inf` accuracy.

inducer: > To me, this seems almost impossibly difficult to determine.

*What* seems almost impossibly difficult to determine?

thomasgibson: That sounds like a good idea to me @alexfikl!

@inducer I guess I should have been more clear; what I am talking about is if someone instantiates an instance of `modepy.Quadrature` by passing in some set of weights and nodes. Given those two bits of info, _what is the order of accuracy_? What I then suggested is:

> ... `modepy.Quadrature` should not be allowed to be instantiated on its own (i.e. make it an abstract class); rather have only concrete subclass of it

That's one proposed solution, which is in line with what @alexfikl is suggesting about having a concrete type for 0-D quadrature. That way, we don't have to play guessing games to figure out what `exact_to` should do.

Granted, this is not a huge problem for most rules. But since we want to check the order of accuracy of the quadrature rule when assembling the face mass matrix, we should make sure we can always get that information in a clear, simple way.

That's it.

alexfikl: To give credit, I was just echoing what @inducer said before

> Also, it might be worthwhile to centralize/standardize that 0D quadrature rule in modepy, so that we don't create multiple inconsistent versions of it.

I'm also ok with `mp.Quadrature` not being an abstract class and just acting as a bag of nodes and weights that the user promises is a quadrature rule. Similar to the `PolynomialGivenNodesElementGroup` thing in meshmode.

inducer: Thanks for explaining!

> That way, we don't have to play guessing games to figure out what `exact_to` should do.

I agree that we shouldn't try to guess `exact_to` from provided nodes and weights, it should be supplied, by someone.

* For the 0D quadrature rule, making it its own class will neatly solve this issue. The class can define an `exact_to` attribute, and we're done.
* For other quadratures, I think it's important to keep apart the notion of exactness, and whether the rule exists as a separate class. It's very possible to make a quadrature subclass that receives `exact_to` data via its constructor argument.
* Unrelated: https://github.com/inducer/modepy/issues/31

grudge/symbolic/operators.py
```diff
@@ -651,20 +658,53 @@ def matrix(self, afgrp, volgrp, dtype):
...
+            else:
+                # NOTE: This handles the general case where
+                # volume and surface quadrature rules may have different
+                # integration orders
+                face_quadrature = mp.quadrature_for_space(
+                    mp.space_for_shape(face, 2*max(n, m)),
+                    face
+                )
+
+            assert face_quadrature.exact_to >= n + m
```
thomasgibson, line 685: @inducer I have a couple questions that's puzzling me and I want to make sure I understand  what's going on.

1. When overintegrating, why is `volgrp` from the native (not overintegrated) discretization? Is it to ensure you have access to the volume basis functions so you can sample at the higher-order quadrature grid?

2. What should the accuracy check be when `afgrp` is coming from a higher-order quadrature grid? I think this assert here:
```python
assert face_quadrautre.exact_to >= n + m
```
Makes sense for all typical cases (no overintegration). No questions there. However,
for example in [this mirgecom test](https://github.com/illinois-ceesd/mirgecom/blob/2a39aa880ff883774f9646e23746c682c151fd80/test/test_diffusion.py#L271), `volgrp.order` is 2, but afgrp.order is 6. The corresponding quadrature rule associated with `afgrp` has an accuracy such that it exactly integrates polynomials of order 6 (based on checking `quad.exact_to`.

So that will fail the assert (at least in 3D). I guess my question is: can't we just safely use the quadrature rule associated with the `QuadratureSimplexElementGroup` and `assert quad.exact_to >= m`? (Otherwise why do we carry around quadrature rules in the first place?)

inducer: > Is it to ensure you have access to the volume basis functions so you can sample at the higher-order quadrature grid?

Yes! Those are the basis functions you need to test against after all.

> can't we just safely use the quadrature rule associated with the `QuadratureSimplexElementGroup` and `assert quad.exact_to >= m`? (Otherwise why do we carry around quadrature rules in the first place?)

That works for me. Possibly, an argument could be made for `quad.exact_to >= 2*m`, but I'm OK with whichever you choose.

Side point: Don't use `assert` for validating things that are under user control. (those should be `if: raise`). `assert` is for things you are able to prove true based on assumptions the code makes and the algorithms you're using. (It's possible that you'll spot bits of the code that don't follow that guideline. If so, feel free to change them.)

inducer:
LGTM! Thanks for seeing this through.

PR #67, opened 2021-04-15
Deprecate grudge.eager in favor of grudge.op

inducer:
- Before you ask: This branch should still run all of mirgecom unmodified. (see the CI)
- See [here](https://github.com/inducer/grudge/commit/7197ad3762fb14241e87eea11d8058d09c244c5f#diff-744059a765032fbe81a80c99e9263537af9cce120787d5d3944cd9e12fdf5528) for the changes in `wave-eager.py` (now `wave-op.`py`) that would be involved.
    - Most changes are of the type `discr.inverse_mass(...)` -> `op.inverse_mass(discr, ...)`.
- Why? `grudge.eager`doesn't make a great deal of sense for what will become *the* interface for grudge (once the current symbolic `sym`/`bind` stuff under the hood goes away). Likewise for `EagerDGDiscretization`. That class name doesn't make much sense when it is used to drive computations with the lazy array context.
- Functions make more sense than methods for most operations. The existing interface was also internally inconsistent: `interior_trace_pair` was already a function, but `inverse_mass` was a method...?
- This also contains a rename of `weak_div` and `weak_grad` to `weak_local_div` and `weak_local_grad`, to help avoid misconceptions.

### DIFF
--- a/examples/advection/surface.py
+++ b/examples/advection/surface.py
@@ -148,8 +148,8 @@ def main(ctx_factory, dim=2, order=4, product_tag=None, visualize=False):
         quad_tag_to_group_factory[product_tag] = \
                 QuadratureSimplexGroupFactory(order=4*order)
 
-    from grudge import DGDiscretizationWithBoundaries
-    discr = DGDiscretizationWithBoundaries(actx, mesh,
+    from grudge import DiscretizationCollection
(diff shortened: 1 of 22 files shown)


inducer:
Additional bait and switch: This renames `DGDiscretizationWithBoundaries` -> `DiscretizationCollection`.

PR #70, opened 2021-04-17
Enable comm of 2d arrays in cross_rank_pairs

MTCam:
Upon detection of an object array *vec* input to `cross_rank_trace_pairs`, flatten the input array in order to use the existing infrastructure to communicate the data without modification, then recover the original shape upon formation of the returned TracePairs.

The current implementation "works" for _MIRGE-Com_'s purpose of communicating the 2D array of grad(Q) naturally via a call to `cross_rank_trace_pairs` in [NS PR](https://github.com/illinois-ceesd/mirgecom/pull/298).

### DIFF
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -478,9 +478,9 @@ def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
 
 
 def cross_rank_trace_pairs(dcoll, ary, tag=None):
-    r"""Get a list of *vec* trace pairs for each partition boundary.
+    r"""Get a list of *ary* trace pairs for each partition boundary.
 
-    For each partition boundary, the field data values in *vec* are
+    For each partition boundary, the field data values in *ary* are
     communicated to/from the neighboring partition. Presumably, this
     communication is MPI (but strictly speaking, may not be, and this
     routine is agnostic to the underlying communication, see e.g.
@@ -489,15 +489,15 @@ def cross_rank_trace_pairs(dcoll, ary, tag=None):
     For each face on each partition boundary, a :class:`TracePair` is
     created with the locally, and remotely owned partition boundary face
     data as the `internal`, and `external` components, respectively.
-    Each of the TracePair components are structured like *vec*.
+    Each of the TracePair components are structured like *ary*.
 
-    The input field data *vec* may be a single
+    The input field data *ary* may be a single
     :class:`~meshmode.dof_array.DOFArray`, or an object
     array of ``DOFArray``\ s of arbitrary shape.
     """
-    if isinstance(vec, np.ndarray):
-        oshape = vec.shape
-        comm_vec = vec.flatten()
+    if isinstance(ary, np.ndarray):
+        oshape = ary.shape
+        comm_vec = ary.flatten()
 
         n, = comm_vec.shape
         result = {}
@@ -520,7 +520,7 @@ def cross_rank_trace_pairs(dcoll, ary, tag=None):
                 )
             for remote_rank in connected_ranks(dcoll)]
     else:
-        return _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=tag)
+        return _cross_rank_trace_pairs_scalar_field(dcoll, ary, tag=tag)
 
 # }}}
 
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -500,7 +500,7 @@ def cross_rank_trace_pairs(dcoll, vec, tag=None):
     boundary.
     """
     if isinstance(vec, np.ndarray):
-        original_shape = vec.shape
+        orig_shape = vec.shape
         comm_vec = vec.flatten()
 
         comm_n, = comm_vec.shape
@@ -512,17 +512,13 @@ def cross_rank_trace_pairs(dcoll, vec, tag=None):
                 assert isinstance(rank_tpair.dd.domain_tag.tag, BTAG_PARTITION)
                 result[rank_tpair.dd.domain_tag.tag.part_nr, ivec] = rank_tpair
 
-        pb_tpairs = []
-        for remote_rank in connected_ranks(dcoll):
-            int_result = make_obj_array([result[remote_rank, i].int
-                                         for i in range(comm_n)])
-            ext_result = make_obj_array([result[remote_rank, i].ext
-                                         for i in range(comm_n)])
-            pb_tpairs.append(TracePair(
-                dd=sym.as_dofdesc(sym.DTAG_BOUNDARY(BTAG_PARTITION(remote_rank))),
-                interior=int_result.reshape(original_shape),
-                exterior=ext_result.reshape(original_shape)))
-        return pb_tpairs
+        return [TracePair(
+            dd=sym.as_dofdesc(sym.DTAG_BOUNDARY(BTAG_PARTITION(remote_rank))),
+            interior=make_obj_array([result[remote_rank, i].int
+                                     for i in range(comm_n)]).reshape(orig_shape),
+            exterior=make_obj_array([result[remote_rank, i].ext
+                                     for i in range(comm_n)]).reshape(orig_shape))
+                for remote_rank in connected_ranks(dcoll)]
     else:
         return _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=tag)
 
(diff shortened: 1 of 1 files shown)


grudge/op.py
```diff
@@ -478,13 +478,36 @@ def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
...
+    """Get a list of *vec* trace pairs for each partition boundary.
+
+    For each partition boundary, the field data values in *vec* are
+    communicated to/from the neighboring partition. Presumably, this
+    communication is MPI (but strictly speaking, may not be, and this
+    routine is agnostic to the underlying communication, see e.g.
+    _cross_rank_trace_pairs_scalar_field).
+
+    For each face on each partition boundary, a :class:`TracePair` is
+    created with the locally, and remote owned partition boundary face
```
inducer, line 490:
```suggestion
    created with the locally, and remotely owned partition boundary face
```

grudge/op.py
```diff
@@ -478,13 +478,36 @@ def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
...
+    data as the `internal`, and `external` components, respectively.
+    Each of the TracePair components are structured like *vec*.
+
+    The input field data *vec* may be a single DOFArray, or an object
```
inducer, line 494:
```suggestion
    The input field data *vec* may be a single
    :class:`~meshmode.dof_array.DOFArray`, or an object
```

grudge/op.py
```diff
@@ -478,13 +478,36 @@ def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
 
 
 def cross_rank_trace_pairs(dcoll, vec, tag=None):
+    """Get a list of *vec* trace pairs for each partition boundary.
```
inducer, line 481:
```suggestion
    r"""Get a list of *vec* trace pairs for each partition boundary.
```

grudge/op.py
```diff
@@ -478,13 +478,36 @@ def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
...
+    boundary.
+    """
     if isinstance(vec, np.ndarray):
+        oshape = vec.shape
+        comm_vec = vec.flatten()
 
-        n, = vec.shape
+        n, = comm_vec.shape
         result = {}
         for ivec in range(n):
```
inducer, line 508:
```suggestion
        # FIXME: Batch this communication rather than
        # doing it in sequence.
        for ivec in range(n):
```

grudge/op.py
```diff
@@ -478,13 +478,36 @@ def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
...
+    data as the `internal`, and `external` components, respectively.
+    Each of the TracePair components are structured like *vec*.
+
+    The input field data *vec* may be a single DOFArray, or an object
+    array of DOFArrays. Each of *vec* components are independently
+    communicated by calling this routine. Upon entry, *vec* is
+    serialized (if needed), each component is communicated, then
+    (if needed) the components are de-serialized back to the original
+    structure of *vec*, before returned as TracePairs for each partition
+    boundary.
```
inducer, line 500:
```suggestion
    array of ``DOFArray``\ s of arbitrary shape.
```
(I don't think it matters to users exactly how the documented effect is achieved, so I don't think we should specify it, especially when the implementation, as it is here, is a tad *cough* suboptimal.)

grudge/op.py
```diff
@@ -478,13 +478,36 @@ def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
 
 
 def cross_rank_trace_pairs(dcoll, vec, tag=None):
```
inducer, line 480:
```suggestion
def cross_rank_trace_pairs(dcoll, ary, tag=None):
```
? (Given that `vec` is a bit of a misnomer now?)

inducer (commented):
Thanks! This looks good already. A few minor nits below.

grudge/op.py
```diff
@@ -477,14 +477,35 @@ def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
...
+    """
+    if isinstance(ary, np.ndarray):
+        oshape = ary.shape
+        comm_vec = ary.flatten()
 
-        n, = vec.shape
+        n, = comm_vec.shape
         result = {}
+        # FIXME: Batch this communication rather than
+        # doing it in sequence.
```
MTCam, line 505: Just note to selves; it may be worthwhile leaving the option to chunk into the smallest (reasonably) possible message.

inducer: I thought "bigger is better" when it comes to MPI comm?

inducer:
> "Apply suggestions from code review" in the commit message

Look at [this view](https://github.com/MTCam/grudge/commits/cross-rank-pairs-2d). To my mind, commit messages serve to help folks figure out what a certain commit was about. In my view, that message, while convenient to type, doesn't totally succeed at that goal.

grudge/op.py
MTCam: Me too, but my experience is based entirely on CPU-type experience.  I thought we had heard recently from Shelby that message latency was getting quite high for even modestly sized messages, potentially favoring smaller messages. I could be mistaken about that 🤷‍♂️

MTCam:
> > "Apply suggestions from code review" in the commit message
> 
> Look at [this view](https://github.com/MTCam/grudge/commits/cross-rank-pairs-2d). To my mind, commit messages serve to help folks figure out what a certain commit was about. In my view, that message, while convenient to type, doesn't totally succeed at that goal.

Sure.  Sorry about that; in fact it is the default message when you use the suggestion batching option, no typing necessary.

grudge/op.py
inducer: True... but I'd regard all those cases as bugs in the MPI implementation. @sll2, would you agree?

inducer:
LGTM. Thanks!

grudge/op.py
majosm: Even if we keep sending separate messages for each component, I imagine it would be preferable to fire them all off and then wait once instead of waiting on each component sequentially, right?

inducer: > Even if we keep sending separate messages for each component, I imagine it would be preferable to fire them all off and then wait once instead of waiting on each component sequentially, right?

Yep, for sure.

MTCam: > Even if we keep sending separate messages for each component, I imagine it would be preferable to fire them all off and then wait once instead of waiting on each component sequentially, right?

As opposed to repeated calls to something like (send/recv wait, send/recv wait, ... ), definitely would be better to do (send/recv, send/recv, wait), yeah. Unless we bump up against some data transfer resource limitation due to how much data we have in simultaneous flight.

sll2: When to batch communication is motivated by the dominating term in the Max-Rate Model. If we're already incurring the injection rate bandwidth of the NIC, then the difference in time required for every process to send 10^6 Bytes of data vs every process sending double that amount appears to be small. 

More information:
When the entire amount of data being injected by a node is large (~10^5 Bytes or higher on Lassen), it's faster to send the data when it's split up evenly across all active processes on a node. This is fairly consistent across machines (Blue Waters, Lassen, Summit) and MPI implementations (spectrum-mpi, mvapich2), so it's not really an artifact of the MPI implementation.

Here's an example below for Lassen, where the Message size is being split up across various processes per node, and the fastest time is achieved by splitting that data sent evenly across all available cores (PPN 40).
[lassen_cpu_node_pong.pdf](https://github.com/inducer/grudge/files/6336983/lassen_cpu_node_pong.pdf)

In our case, however, presumably the data being sent off node is already split evenly across the active processes.

MTCam: Just for context...

At each RK substep of our production problem, we need to exchange ~500Bytes total per partition boundary (PB) point.  This includes exchanging only **Q** and grad(**Q**) - safe for now.

That translates to about (2KB, 3KB, 5KB) per boundary triangle for order (1, 2, 3) triangles, respectively.  The lower bound (i.e. the (reasonably) smallest) message size is that where we bust it into one message per scalar component so ~(24B, 50B, 100B) per boundary triangle.

The total message size depends on the number of triangles on each PB, which is clearly a function of mesh resolution and partitioning. For say 100 order 1 triangles on each PB that means the smallest possible message size for order 1 runs is ~3KB, and big message of ~20KB for a (fully) batched exchange.

PR #71, opened 2021-04-19
Add jump property to TracePair.

MTCam:
No problem to just write `pair.ext - pair.int`, but adding this does make the expressions in _MIRGE-Com_ nicer to read.


inducer:
Heh, there's actually a reason why `avg` is there, but `jump` is not. The reason is that it's not *entirely* obvious what "jump" should be.

The closest I've seen to a consensus definition in the literature is the following:
grafik

(Before Sec. 3.2 in https://doi.org/10.1137/S0036142901384162) By that, the jump of a scalar quantity is vectorial, and the jump of a vectorial quantity is scalar. That's basically the only definition that I would be willing to go along with, but it's still packaging up a lot of things, which is why I've held off so far.

cc @lukeolson @thomasgibson @majosm Opinions?

majosm:
Maybe if `jump` is confusing, we could call it something else? `diff`?

MTCam:
> Maybe if `jump` is confusing, we could call it something else? `diff`?
> 
> ~Edit: Do you have any usage examples @MTCam?~ Never mind, I think I can see the use. So, in grudge notation what @inducer posted would look something like `tpair.jump*normal` (scalar case) or `dot(tpair.jump, normal)` (vector case)?

Yes. By _jump_, I just meant the signed difference between the pair quantities. I think the excerpt (from what looks like Hesthaven) given by @inducer is referring to the action of the so-called jump operator `[[.]]` used by the authors of that reference.

MTCam:
> Hmm, `ext - int` or `int - ext`? 🙂

I like `ext - int` only because in the literature on shocks, jump conditions, the "jump" in X is defined as (X2 - X1) where X1 is X(at -infinity) to the left of the shock, and X2 is X to the right, which I psychologically associate with (int), (ext), respectively.

MTCam:
See e.g. this excerpt from [Jameson's](https://www.researchgate.net/profile/Antony-Jameson/publication/24293067_Analysis_And_Design_Of_Numerical_Schemes_for_Gas_Dynamics_2_Artificial_Diffusion_and_Discrete_Shock_Structure/links/545ec9530cf2c1a63bfc2193/Analysis-And-Design-Of-Numerical-Schemes-for-Gas-Dynamics-2-Artificial-Diffusion-and-Discrete-Shock-Structure.pdf):

Screen Shot 2021-04-19 at 3 06 17 PM

majosm:
Only downside of `ext - int` is that when implementing the jump operators you have to remember to add a minus sign, because e.g. `q_int*n_int + q_ext*n_ext = q_int*n_int + q_ext*(-n_int) = (q_int-q_ext)*n_int = -diff*n_int`. Not sure how big of a deal that is.

inducer:
Given all these different possibilities, I'm really having trouble picking a winner...

MTCam:
> Only downside of `ext - int` is that when implementing the jump operators you have to remember to add a minus sign, because e.g. `q_int*n_int + q_ext*n_ext = q_int*n_int + q_ext*(-n_int) = (q_int-q_ext)*n_int = -diff*n_int`. Not sure how big of a deal that is.

That isn't a downside from my perspective, and it pleases me to be able to directly correlate the widely used convention (and as used by Jameson above) with the expression in the code.

thomasgibson:
I think `diff = ext - int` is pretty good from an element-wise perspective. The definition presented by @inducer above makes perfect sense when you are talking about the jump w.r.t. facet-wise iteration. But since grudge is primarily concerned with element-wise operations (meaning your FEM kernels are performed cell-wise, not facet-wise), then `diff = ext - int` is consistent!

(unless I'm totally wrong and grudge emits loopy code that runs over facets for doing surface integrals?)

inducer:
@thomasgibson That's a fair point, I hadn't looked at it from that angle. You're right in that there always is a notion of a "home/local cell". I'm warming to this.

inducer:
@MTCam Could you introduce some places (in examples and/or tests) where this is used, so that it actually gets used?

MTCam:
> @MTCam Could you introduce some places (in examples and/or tests) where this is used, so that it actually gets tested?

Specifically, I can use it in expressions for numerical fluxes involving the jump at the boundary because as you note the jump in a quantity [.] _along a normal_ is defined as [u]n_hat for scalar u and [u].n for vector u.  I currently have [this placeholder in NS](https://github.com/illinois-ceesd/mirgecom/blob/d388c829ed07fcd96ebc678e94d726c58e1e7a8f/mirgecom/operators.py#L45).

For the use-case in LFR numerical flux: h = fi_tpair.avg@normal - lambda*q_tpair.diff/2

I would switch it up to use diff as soon as it lands if OK.

inducer:
If you're still interested in this, what I meant was: could you modify some of the examples here to use this facility, too?

MTCam:
> If you're still interested in this, what I meant was: could you modify some of the examples here to use this facility, too?

Ugh, in addition to being dense, I'm just not that creative.  (f51930a)

inducer:
Thanks, LGTM! (Tweaked the docstrings a bit in 7ddcaf2.)

PR #73, opened 2021-04-20
Add pylint

alexfikl:
Found some more small typos, so went ahead and threw pylint at it and fixed them. Gitlab CI is running at
https://gitlab.tiker.net/fikl2/grudge/-/pipelines/154607

There's a big chunk in `models` that's ignored, but those don't seem to be up to date anyway.


inducer:
Thanks!

Issue #16, opened 2020-08-29
Eager nodal reductions should respect MPI


inducer:
cc @Mtcam

thomasgibson:
Have we sketched out what needs to be done to actually do this? Right now nodal reductions are rank-local. So could we perhaps introduce something similar to [this](https://github.com/inducer/grudge/blob/main/grudge/op.py#L457-L497), but for volumes?

Would this require extending `DTAG_VOLUME_ALL` to something like `DTAG_VOLUME_PARTITION(rank_id)` or something like that?

inducer:
> Would this require extending `DTAG_VOLUME_ALL` to something like `DTAG_VOLUME_PARTITION(rank_id)` or something like that?

I actually don't hate that. We definitely want to retain the ability to ask for rank-local reductions. One possible issue is that we probably also want to be able to ask for `DTAG_VOLUME_PARTITION(rank=12) & DTAG_VOL_SCRAMJET_WALL`. (before you panic: not arbitrary intersections, just one rank and one domain subset)

cc @majosm because this very much interacts with the subdomain stuff he is building.

PR #74, opened 2021-04-22
New world Grudge

thomasgibson:
This PR revamps Grudge by completely deprecating the symbolic infrastructure for the core DG operators. Instead, core DG routines will call out directly to loopy programs without going through the extensive symbolic pipeline.

All examples and tests have been converted over (with a few exceptions).

The symbolic code will remain for now, as there are a few isolated examples (`dagrt-fusion.py` for example) that still use it, as well as some external functions (issue for reference: #93)


inducer (commented):
This was pending from during our discussion.

grudge/dof_desc.py
```diff
@@ -0,0 +1,223 @@
+"""
+DOF description
```
inducer, line 2: Maybe the `dof_desc` move should be a separate PR. That's something we could merge right away.

thomasgibson: Good idea, I'll create a PR for this

thomasgibson: @inducer I've created the PR here: https://github.com/inducer/grudge/pull/75

Note that after that gets merged, I will rebase this branch on top of that, which means I will be changing some of the commit history so the transition is cleaner (in particular, I will edit this commit: https://github.com/inducer/grudge/pull/74/commits/445e560a93ad840d55e8929f63982864d7a25b9b)

grudge/op.py
```diff
@@ -310,6 +326,109 @@ def mass(dcoll, *args):
...
+    area_elements = area_element(actx, dcoll, dd=dd_in)
+
+    data = []
+    # FIXME: I think there is a bug in loopy's implementation
+    # of einsum; this code works when dispatching to numpy. But
+    # fails when using actx.einsum due to strides mismatch
+    # on the mass matrix... (error raised when attempting to
+    # invoke the loopy kernel)
+    # NOTE: This ONLY fails when in_grp != out_grp
+    # TODO: Need to debug the loopy program...?
```
thomasgibson, line 376: @inducer I got the metric computations working! :tada:

But I think there is something wonky going on when using loopy's `einsum` implementation. I'm unable to pass the test in `test_new_world_grudge.py` (I get something like: `TypeError: strides mismatch on argument 'mass_mat' (after removing unit length dims...`).

It does, however, pass when I dispatch to `numpy.einsum` (as in the code down below shows). I rewrote the mass matrix routine to make it more debuggable.

I think I managed to track down where it's breaking; the exception is getting generated [here](https://github.com/inducer/loopy/blob/main/loopy/target/execution.py#L547), but I'm afraid it will take me some time to decipher this bit of loopy... I'll revisit this tomorrow.

inducer: Unlike numpy, loopy-generated code is picky about the memory layout (strides) of the array it is given. (If you think about it, it *has* to be, given that it's generating C code!) You can manipulate this using [`tag_array_axes`](https://documen.tician.de/loopy/ref_transform.html?highlight=tag_array_axes#loopy.tag_array_axes), specifically. By default, loopy expects C-layout (row-major) arrays. Your matrix likely comes out of LAPACK and is therefore in Fortran (column-major) layout.

thomasgibson: @inducer Thanks for the info. I'll try to experiment with it.

BTW I forgot to mention that the test only fails when `in_grp != out_grp`. IOW: when the mass matrix is not square due to using a quadrature element group as the `in_grp`.

grudge/op.py
```diff
@@ -310,6 +326,82 @@ def mass(dcoll, *args):
...
+                              out_element_group=out_grp,
+                              in_element_group=in_grp
+                          ),
+                          ae_i,
+                          vec_i,
+                          arg_names=("mass_mat", "jac_det", "vec"),
+                          tagged=(MassOperatorTag(),),
+                          tagged_array_axes={
+                              "mass_mat": "stride:auto,stride:auto"
+                           })
```
thomasgibson, line 380: @inducer I managed to get the mass matrix test to finally work. Requires: https://github.com/inducer/meshmode/pull/170

I'm not 100% sure this is the best way to do this, as this was a rather quick attempt to get something working. To be perfectly honest, I'm still not 100% sure if there are any unknown consequences for choosing these parameters.

I read up the docs [here](https://documen.tician.de/loopy/ref_transform.html?highlight=tag_array_axes#loopy.tag_array_axes) but they weren't extremely illuminating. I dug around a bit in the code [here](https://github.com/inducer/loopy/blob/main/loopy/kernel/array.py#L210-L417) (the docs suggested [here](https://documen.tician.de/loopy/ref_internals.html#loopy.kernel.array.parse_array_dim_tags) but there's nothing there, so I went straight to the source code) and it was even less obvious. I didn't see a summary describing the acceptable parameters, so I dug line-by-line to back out what acceptable parameters were (as well as taking some hints from some examples in the test suite).

Maybe I wasn't looking in the right place.

inducer: Remind me sometime to show you how to look at the generated CL code, to make all this a bit less mysterious. (In fact, I'll just add that to our agenda for tomorrow.)

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,256 @@
+
```
inducer, line 1: All files need a copyright/license header.

grudge/geometry/__init__.py
```diff
@@ -0,0 +1,2 @@
+
+from grudge.geometry.metrics import *
```
inducer, line 2: Link into documentation? That'll also determine the "canonical" import spot for these things.

inducer, line 2: Not a huge fan on `*` imports... could you maintain the list?

test/test_new_world_grudge.py
```diff
@@ -0,0 +1,367 @@
+__copyright__ = """
```
inducer, line 1: This file name won't make a great deal of sense once old-world grudge goes away.

test/test_new_world_grudge.py
```diff
@@ -0,0 +1,367 @@
...
+Copyright (C) 2021 University of Illinois Board of Trustees
```
inducer, line 2: I'm guessing you're already doing this, but I'd like it agreed upon as a policy: What I'd like here to make reviewing easier is to have a file where there's a 1-to-1 correspondence between new-world and old-world tests, to make sure we don't lose coverage.

grudge/geometry/__init__.py
inducer, line 2: Not specific to this, but a general comment: For review, I'm kind of wary of letting this thing grow too big. I'd be more in favor of merging it incrementally.

grudge/op.py
```diff
@@ -145,6 +161,56 @@ def local_grad(dcoll, vec):
     return _bound_grad(dcoll)(u=vec)
 
 
+def reference_derivative_matrices(actx, element_group):
```
inducer, line 164: This file is getting big. Are you OK with adding sectioning comments?

grudge/op.py
```diff
@@ -145,6 +161,56 @@ def local_grad(dcoll, vec):
...
+        )
+    return DOFArray(actx, data=tuple(data))
+
+
+def local_gradient(dcoll, vec):
+    from pytools.obj_array import make_obj_array
+    return make_obj_array(
+        [_compute_local_gradient(dcoll, vec, xyz_axis)
+         for xyz_axis in range(dcoll.dim)]
+    )
```
inducer, line 211: Once you postmultiply with the chain rule geometric factors, it might be smart to measure the performance impact. We might need to put in place a hand-coded 3x3 loopy "matvec" to avoid regressing eager performance more (and only activate that for the eager case).

inducer (commented):
Thanks for taking on this task! A few notes below.

inducer:

grudge/geometry/__init__.py
inducer: > Would the next logical thing be to split out the geometry computations?

Yeah, I think that would make sense.

test/test_new_world_grudge.py
thomasgibson: `test_new_world_grudge.py ` is not intended to remain in the long term. I want to gradually port over tests as more functionally is converted and test them.

grudge/op.py
thomasgibson: I split out the new operators into a separate `operators.py` file, and I've added sectioning comments there too

grudge/op.py
thomasgibson: @inducer Could you elaborate a little more what you're suggesting?

I agree with you 100% we should monitor performance! Are you suggesting that we may want to just compute the gradients "all at once" via a hand-coded loopy program?

thomasgibson:
There are currently some memory allocation issues being raised in the CI... need to track down what the heck is going on there.

grudge/op.py
```diff
@@ -193,16 +259,102 @@ def local_div(dcoll, vecs):
...
+    # FIXME: Is there a better/more efficient way to do this?
+    inverse_jac_t = actx.from_numpy(
+        np.asarray(
+            [actx.to_numpy(
+                inverse_surface_metric_derivative(
+                    actx, dcoll, rst_axis, xyz_axis, dd=dd_in
+                )[0]
+            ) for rst_axis in range(dcoll.dim)]
+        )
+    )
```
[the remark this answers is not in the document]
thomasgibson: Used your suggestion on slack: `actx.np.stack`!

test/test_grudge.py
```diff
@@ -1109,47 +1096,77 @@ def test_norm_obj_array(actx_factory, p):
...
+    normal = op.normal(dcoll, BTAG_NONE)
+    from meshmode.dof_array import DOFArray
+    for component in normal:
+        assert isinstance(component, DOFArray)
+        assert len(component) == len(dcoll.discr_from_dd(BTAG_NONE).groups)
 
-    sym_if = sym.If(sym.Comparison(2.0, "<", 1.0e-14), 1.0, 2.0)
-    bind(discr, sym_if)(actx)
 
+# {{{ Deprecated symbolic tests
```
thomasgibson, line 1118: Symbolic specific tests which do not have a "new world grudge" equivalent are moved down here

test/test_grudge.py
```diff
@@ -950,35 +970,6 @@ def conv_test(descr, use_quad):
...
-
-    v1 = sym.var("v1")
-    ob1 = sym.OperatorBinding(TestOperator(), v1)
-
-    # The output order isn't significant, but it should always be the same.
-    assert list(TestBoundOperatorCollector(TestOperator)(ob0 + ob1)) == [ob0, ob1]
-
-# }}}
-
-
```
thomasgibson, line 981: Moved under `# {{{ Deprecated symbolic tests`

test/test_grudge.py
```diff
@@ -991,105 +982,102 @@ def test_bessel(actx_factory):
...
+    nodes = thaw(actx, op.nodes(dcoll))
+    r = actx.np.sqrt(nodes[0]**2 + nodes[1]**2)
 
-    nodes = sym.nodes(dims)
-    r = sym.cse(sym.sqrt(nodes[0]**2 + nodes[1]**2))
+    # FIXME: Bessel functions need to brought out of the symbolic
+    # layer. Related issue: https://github.com/inducer/grudge/issues/93
+    def bessel_j(actx, n, r):
+        from grudge import sym, bind
+        return bind(dcoll, sym.bessel_j(n, sym.var("r")))(actx, r=r)
```
thomasgibson, line 994: @inducer I've left this test partially converted, but still call out to `sym` to evaluate the Bessel functions. I've linked the related issue #93 here for reference

thomasgibson:
@inducer Are you okay with me completely removing most of `grudge/models` (`gas_dynamics`, `burgers.py`, `diffusion.py`, `nd_calculus.py`, and `pml.py` specifically)? The models: `advection.py`, `em.py` and `wave.py` are fully converted and exercised in the examples/unit tests.

What about the files in `examples`: `dagrt-fusion.py` and `geometry.py`? `geometry.py` is a simple enough script to port, but `dagrt-fusion.py` is _way_ more reliant on the symbolics. I am leaning towards completely removing it. Any reason why we should not? And if not, the only other solution I can think of is to keep a "symbolic wave operator" floating around so `dagrt-fusion.py` can use that.

inducer:
For `dagrt-fusion.py`, maybe just copy the wave operator to live alongside it. Maybe move it to a "old-symbolic" subfolder?

grudge/op.py
inducer: We're not bothering with manual fusion anywhere else, so let's not start here. I take back what I said.

grudge/geometry/__init__.py
thomasgibson: I've updated the geometry module and added documentation throughout. They should be included in the main doc build!

grudge/op.py
```diff
@@ -403,155 +995,133 @@ def norm(dcoll, vec, p, dd=None):
...
-                interior=make_obj_array([
-                    result[remote_rank, i].int for i in range(n)]).reshape(oshape),
-                exterior=make_obj_array([
-                    result[remote_rank, i].ext for i in range(n)]).reshape(oshape)
-                )
-            for remote_rank in connected_ranks(dcoll)]
-    else:
-        return _cross_rank_trace_pairs_scalar_field(dcoll, ary, tag=tag)
+    if isinstance(vec, np.ndarray):
+        return sum(elementwise_sum(dcoll, dd, vec_i) for vec_i in vec)
```
thomasgibson, line 1103: @inducer Actually, looking at this now, I think this should be
```python
    if isinstance(vec, np.ndarray):
        return obj_array_vectorize(
            lambda vi: elementwise_sum(dcoll, dd, vi), vec
        )
```
What do you think?

thomasgibson: I'm pretty sure this is what it should be. So I'm going to add it in.

thomasgibson:
@inducer I probably jumped too quickly to adopt https://github.com/inducer/grudge/pull/74/commits/e27c40bbefc3c506dca0a3bc7918ef12019a343c before https://github.com/inducer/pyopencl/pull/457 was added to the conda archives. Errors in CI are related to that.

inducer:
Generally, I have to roll a release of pyopencl in order for the CI to see updates of pyopencl. I did that, based on a request from @kaushikcfd, so this should be OK now. (though it'll build pyopencl from source, see https://github.com/inducer/ci-support/pull/15/commits/99f7fb9ae98c55c987c2e647aabe8ffa41418b9a)

inducer:
Reran the CI, there's a new mirgecom failure implicating `maximum`. Could you take a look?

thomasgibson:
@inducer Looks like, for whatever reason, the downstream test used pyopencl v 2021.2.1 https://github.com/inducer/grudge/pull/74/checks?check_run_id=2647253813#step:3:759 (instead of 2021.2.2). I'm going to rerun to see if it's just a hiccup.

grudge/op.py
```diff
@@ -403,155 +998,147 @@ def norm(dcoll, vec, p, dd=None):
...
-    return _nodal_reduction(dcoll, sym.NodalMax, dd)(arg=vec)
+def nodal_summation(vec):
+    r"""Return the nodal sum of a vector of degrees of freedom *vec*.
 
-# }}}
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
+    :returns: an integer denoting the nodal sum.
+    """
+    actx = vec.array_context
+    return sum([actx.np.sum(grp_ary) for grp_ary in vec])
```
thomasgibson, line 1018: @inducer I'm not rethinking this choice (similarly for `op.nodal_{minimum, maximum}`). I think we will probably want to keep `dcoll` as a passed argument. The reason being is that I'm thinking ahead for MPI-aware nodal reductions (https://github.com/inducer/grudge/issues/16) and we'll definitely need the discretization collection for doing cross-rank reductions. What do you think?

inducer: I agree with keeping `dcoll` as an argument.

grudge/op.py
```diff
@@ -47,38 +24,136 @@
...
+
+class HasElementwiseMatvecTag(FirstAxisIsElementsTag):
+    """A tag that is applicable to kernel programs indicating that
+    an element-wise matrix product is being performed. This indicates
+    that the first index corresponds to element indices and suggests that
+    the implementation should set element indices as the outermost
+    loop extent.
+    """
+
+# }}}
```
thomasgibson, line 132: @inducer Should we move tags to something like `metadata.py`?

examples/old_symbolics/symbolic_wave_op.py
```diff
@@ -0,0 +1,166 @@
+"""(Old world) symbolic wave equation operator."""
```
thomasgibson, line 1: This file is just a plain copy of the previous symbolic wave equation operator. (for `dagrt-fusion.py`)

grudge/op.py
```diff
@@ -47,38 +24,136 @@
...
-from meshmode.mesh import BTAG_ALL, BTAG_NONE, BTAG_PARTITION  # noqa
-from meshmode.dof_array import freeze, flatten, unflatten
+from grudge.trace_pair import (  # noqa
+    interior_trace_pair,
+    interior_trace_pairs,
+    connected_ranks,
+    cross_rank_trace_pairs,
+    bdry_trace_pair,
+    bv_trace_pair
+)
```
thomasgibson, line 54: These are imported here so we can use `grudge.op` as the main source of core DG-related functions. I.e. `op.interior_trace_pairs`, `op.cross_rank_trace_pairs`, etc

grudge/op.py
```diff
@@ -403,155 +1002,153 @@ def norm(dcoll, vec, p, dd=None):
...
+            name="grudge_elementwise_%s_knl" % op_name
+        )
+    return prg()
 
-        bdry_conn = self.dcoll.get_distributed_boundary_swap_connection(
-                dof_desc.as_dofdesc(dof_desc.DTAG_BOUNDARY(self.remote_btag)))
-        swapped_remote_dof_array = bdry_conn(remote_dof_array)
 
-        self.send_req.Wait()
+def elementwise_sum(dcoll, *args):
```
thomasgibson, line 1110: I didn't port over `elementwise_{mix,max}` since it's used literally nowhere in the mirgecom stack (or even internally in old world grudge). However, I believe I've written it in such a way that it would be very easy to add them in later if we need them.

grudge/trace_pair.py
```diff
@@ -0,0 +1,361 @@
...
+    """
+
+    dd: dof_desc.DOFDesc
+    interior: ArrayContainer
+    exterior: ArrayContainer
+
+    def __init__(self, dd, *, interior, exterior):
+        object.__setattr__(self, "dd", dof_desc.as_dofdesc(dd))
+        object.__setattr__(self, "interior", interior)
+        object.__setattr__(self, "exterior", exterior)
```
thomasgibson, line 103: This might look crazy at first glance, but you can't do `self.dd = ...` (etc) for frozen `dataclasses`. I wanted to keep the constructor here (hence `init=False` in the `@dataclass` decorator). Ref: https://docs.python.org/3/library/dataclasses.html#frozen-instances

inducer:
Working on this now, just pushed a bunch of changes (not done). Please look them over and comment.

inducer:
You OK with 677d7da?

thomasgibson:
> You OK with [677d7da](https://github.com/inducer/grudge/commit/677d7dab08ff521243d733f3a8ab9dd4e87b51f7)?

Did you mean to drop those tests in the file `test/test_grudge_sym_old.py`? It's currently showing as empty. But yes, I'm totally happy with moving those into a separate file!

inducer:
Good catch! Fixed up in ff5c3ea.

examples/wave/wave-min.py
```diff
@@ -1,7 +1,9 @@
 """Minimal example of a grudge driver."""
```
inducer, line 1: Could you add "does-not-blow-up" asserts to all the examples?

inducer (commented):
Thanks for working on this! In the interest of efficiency, here's a first batch of review comments.

inducer:
(Sorry about those test failures, will take care of them.)

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,712 @@
...
+    # exterior normals for positively oriented curves.
+
+    pder = pseudoscalar(actx, dcoll, dim=dim, dd=dd) \
+        / area_element(actx, dcoll, dim=dim, dd=dd)
+
+    # Dorst Section 3.7.2
+    return pder << pder.I.inv()
+
+
+def mv_normal(actx, dcoll, dd):
```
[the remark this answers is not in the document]
alexfikl: Yeah, that poor normal function is trying to do a lot of work to support `dim - 1` surfaces and `dim - 2` surfaces (i.e. faces of the `dim - 1` surfaces) in a `dim` ambient space. A lot of that is just to get DG on surfaces working, so maybe it's worth having two functions for this
* one that just does the common flat spaces case
* one that does that fancy hypersurfaces case

?

grudge/op.py
```diff
@@ -403,155 +1002,153 @@ def norm(dcoll, vec, p, dd=None):
...
-@memoize_on_first_arg
-def _nodal_reduction(dcoll, operator, dd):
-    return bind(dcoll, operator(dd)(sym.var("arg")), local_only=True)
+    return _norm(dcoll, vec, p, dd)
 
 
 def nodal_sum(dcoll, dd, vec):
-    return _nodal_reduction(dcoll, sym.NodalSum, dd)(arg=vec)
+    from warnings import warn
+    warn("Using 'nodal_sum' is deprecated, use 'nodal_summation' instead.",
```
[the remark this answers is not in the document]
thomasgibson: The only reason I did so is because `dd` is not needed. But I'm rethinking this now and I think it's better to just leave them as is. I'll update this

grudge/op.py
```diff
@@ -331,34 +653,283 @@ def mass(dcoll, *args):
...
+        return make_loopy_program(
+            [
+                "{[iel]: 0 <= iel < nelements}",
+                "{[f]: 0 <= f < nfaces}",
+                "{[idof]: 0 <= idof < nvol_nodes}",
+                "{[jdof]: 0 <= jdof < nface_nodes}"
+            ],
+            """
+            result[iel, idof] = sum(f, sum(jdof, mat[idof, f, jdof]        \
+                                                 * jac_surf[f, iel, jdof]  \
```
inducer, line 878: These backslashes shouldn't be necessary. Just like Python, loopy won't consider a line ended until all parens are closed.

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,710 @@
...
+    return product(
+        forward_metric_derivative_mv(actx, dcoll, rst_axis, dd)
+        for rst_axis in range(dim)
+    )
+
+
+def pseudoscalar(actx: ArrayContext, dcoll: DiscretizationCollection,
+        dim=None, dd=None) -> MultiVector:
+    r"""Computes the field of pseudoscalars for the domain/discretization
+    identified by *dd*
```
inducer, line 464:
```suggestion
    identified by *dd*.
```

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,710 @@
...
+    ).project_max_grade()
+
+
+@memoize_on_first_arg
+def area_element(
+        actx: ArrayContext, dcoll: DiscretizationCollection, dim=None, dd=None
+        ) -> DOFArray:
+    r"""Computes the measure of a transformed element. These are used
+    to transform integrals from global to reference space, and are
+    commonly referred to as Jacobian determinants.
```
inducer, line 490: I find the term "measure" here a bit odd, since it's technically [not one](https://en.wikipedia.org/wiki/Measure_(mathematics)). And the Jacobian determinant bit is only true AFAIK for the volume case (ambient == topological dim). I like the second bit here more: This is the scale factor needed to transform an integral from reference to global.

grudge/op.py
inducer: OK, sure. Maybe revisit for #99.

grudge/op.py
```diff
@@ -48,37 +25,136 @@
...
+
+
+__doc__ = """
+Metadata and geometry
+^^^^^^^^^^^^^^^^^^^^^
+
+Kernel tags
+-----------
+
+.. autoclass:: HasElementwiseMatvecTag
```
inducer, line 65: Not sure I'm ready to document this.

grudge/op.py
inducer: Nah. Not yet... the transform story is still too up in the air.

inducer (commented):
Thanks again for working on this! This is looking great. Below are a few more comments. Once these and the prior batch are dealt with, this is ready to land from my perspective.

inducer:

grudge/discretization.py
```diff
@@ -21,45 +24,64 @@
...
+    .. autoattribute:: real_dtype
+    .. autoattribute:: complex_dtype
 
-    .. autoattribute :: dim
-    .. autoattribute :: ambient_dim
-    .. autoattribute :: mesh
+    .. automethod:: discr_from_dd
+    .. automethod:: connection_from_dds
+    .. automethod:: opposite_face_connection
+    .. automethod:: get_distributed_boundary_swap_connection
```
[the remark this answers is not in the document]
thomasgibson: Fair enough! I'll leave the docstring but remote the doc generation here

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,710 @@
...
+
+    if dim is None:
+        dim = dcoll.discr_from_dd(dd).dim
+
+    return parametrization_derivative(
+        actx, dcoll, dim, dd
+    ).project_max_grade()
+
+
+@memoize_on_first_arg
```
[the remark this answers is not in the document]
thomasgibson: So I just checked, and `hash(dcoll)` does indeed work and returns a hash :thinking:

"Objects which are instances of user-defined classes are hashable by default. They all compare unequal (except with themselves), and their hash value is derived from their id()." source: https://docs.python.org/3/glossary.html#term-hashable

So it's using the address of the object in memory as a hash id. I'll use `@memoize_in` since I think it's safer!

grudge/models/advection.py
```diff
@@ -58,210 +64,312 @@ def advection_weak_flux(flux_type, u, velocity):
...
+                    tpair.dd.with_discr_tag(quad_tag), tpair.int
+                ),
+                exterior=op.project(
+                    dcoll, tpair.dd,
+                    tpair.dd.with_discr_tag(quad_tag), tpair.ext
+                )
+            ) for tpair in op.interior_trace_pairs(dcoll, u)
+        ]
+    else:
+        return op.interior_trace_pairs(dcoll, u)
```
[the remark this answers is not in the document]
thomasgibson: Not out of the box, need to add some extra logic to handle interpolation to quadrature grids. But that should be straightforward enough to add!

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,712 @@
...
+    transformation of physical gradients.
+
+    :arg actx: an :class:`~arraycontext.context.ArrayContext`.
+    :arg dcoll: a :class:`grudge.discretization.DiscretizationCollection`.
+    :arg rst_axis: an integer denoting the reference coordinate axis.
+    :arg xyz_axis: an integer denoting the physical coordinate axis.
+    :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
+        Defaults to the base volume discretization.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` containing the
+        inverse metric derivative at each nodal coordinate.
```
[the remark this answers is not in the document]
thomasgibson: Yep. It's returning a `DOFArray` because `forward_metric_nth_derivative` (which most of everything in `metrics.py` eventually calls) returns a `DOFArray`. So `ndarray` and `MultiVector` containing `DOFArrays` are being passed around.

grudge/op.py
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/d9bb98d696a00f727a4c29166f0debdfd1019b96

grudge/op.py
```diff
@@ -48,37 +25,136 @@
...
+    interior_trace_pair,
+    interior_trace_pairs,
+    connected_ranks,
+    cross_rank_trace_pairs,
+    bdry_trace_pair,
+    bv_trace_pair
+)
+
+
+__doc__ = """
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/d9bb98d696a00f727a4c29166f0debdfd1019b96

grudge/op.py
```diff
@@ -403,155 +992,148 @@ def norm(dcoll, vec, p, dd=None):
...
-# }}}
 
+def nodal_maximum(dcoll: DiscretizationCollection, vec):
+    r"""Return the nodal maximum of a vector of degrees of freedom *vec*.
 
-# {{{ distributed-memory functionality
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
+    :returns: a scalar denoting the nodal maximum.
+    """
+    # FIXME: Make MPI-aware
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/d9bb98d696a00f727a4c29166f0debdfd1019b96

grudge/op.py
```diff
@@ -403,155 +992,148 @@ def norm(dcoll, vec, p, dd=None):
...
+    r"""Return the nodal minimum of a vector of degrees of freedom *vec*.
 
-def interior_trace_pair(dcoll, vec):
-    """Return a :class:`grudge.sym.TracePair` for the interior faces of
-    *dcoll*.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
+    :returns: a scalar denoting the nodal minimum.
     """
-    i = project(dcoll, "vol", "int_faces", vec)
+    # FIXME: Make MPI-aware
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/d9bb98d696a00f727a4c29166f0debdfd1019b96

grudge/op.py
```diff
@@ -403,155 +992,148 @@ def norm(dcoll, vec, p, dd=None):
...
-
-
-def nodal_max(dcoll, dd, vec):
-    return _nodal_reduction(dcoll, sym.NodalMax, dd)(arg=vec)
-
-# }}}
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
+    :returns: a scalar denoting the nodal sum.
+    """
+    # FIXME: Make MPI-aware
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/d9bb98d696a00f727a4c29166f0debdfd1019b96

grudge/op.py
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/d9bb98d696a00f727a4c29166f0debdfd1019b96

examples/advection/surface.py
```diff
@@ -1,4 +1,9 @@
-__copyright__ = "Copyright (C) 2020 Alexandru Fikl"
+"""Minimal example of a grudge driver."""
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/d93c422fecec17f4f6163a84db8a0367413759eb

grudge/geometry/metrics.py
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/3651ce2c3c6f742c5f9a3ac82886a911e58bb188

grudge/geometry/metrics.py
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/3651ce2c3c6f742c5f9a3ac82886a911e58bb188

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,710 @@
...
+    DD_VOLUME, DOFDesc, DISCR_TAG_BASE
+)
+
+from pymbolic.geometric_algebra import MultiVector
+
+from pytools.obj_array import make_obj_array
+from pytools import memoize_on_first_arg
+
+
+__doc__ = """
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/3651ce2c3c6f742c5f9a3ac82886a911e58bb188

grudge/geometry/metrics.py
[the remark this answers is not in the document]
thomasgibson: I've gone ahead and just renamed and made the method "private." We can sort this out in a different PR. https://github.com/inducer/grudge/pull/74/commits/3651ce2c3c6f742c5f9a3ac82886a911e58bb188

Maybe create an issue?

grudge/discretization.py
```diff
@@ -482,28 +536,49 @@ def _all_faces_volume_connection(self):
 
     @property
     def dim(self):
+        """Return the topological dimension."""
         return self._volume_discr.dim
 
     @property
     def ambient_dim(self):
+        """Return the geometric dimension."""
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/9c23f712ee368c7b4524d3a11164937393e0d767

grudge/discretization.py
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/9c23f712ee368c7b4524d3a11164937393e0d767

test/test_grudge.py
```diff
@@ -120,48 +124,50 @@ def test_mass_mat_trig(actx_factory, ambient_dim, discr_tag):
...
-    ones_quad = bind(discr, sym_ones)(actx)
+    quad_disc = dcoll.discr_from_dd(dd_quad)
+    x_quad = thaw(quad_disc.nodes(), actx)
+    f_quad = f(x_quad)
+    ones_quad = quad_disc.zeros(actx) + 1
 
-    mass_op = bind(discr, sym.MassOperator(dd_quad, dof_desc.DD_VOLUME)(sym_f))
+    mop_1 = op.mass(dcoll, dd_quad, f_quad)
+    num_integral_1 = np.dot(actx.to_numpy(flatten(ones_volm)),
+                            actx.to_numpy(flatten(mop_1)))
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/daffc57f4a150368b3dd5370bea014bed7d57471

test/test_grudge.py
```diff
@@ -226,39 +232,41 @@ def test_mass_surface_area(actx_factory, name):
...
 
         # {{{ compute surface area
 
         dd = dof_desc.DD_VOLUME
-        sym_op = sym.NodalSum(dd)(sym.MassOperator(dd, dd)(sym.Ones(dd)))
-        approx_surface_area = bind(discr, sym_op)(actx)
+        ones_volm = volume_discr.zeros(actx) + 1
+        flattened_mass_weights = flatten(op.mass(dcoll, dd, ones_volm))
+        approx_surface_area = np.dot(actx.to_numpy(flatten(ones_volm)),
+                                     actx.to_numpy(flattened_mass_weights))
```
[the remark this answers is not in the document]
thomasgibson: https://github.com/inducer/grudge/pull/74/commits/daffc57f4a150368b3dd5370bea014bed7d57471

grudge/trace_pair.py
```diff
@@ -0,0 +1,361 @@
...
+    def __getattr__(self, name):
+        return map_array_container(
+            lambda ary: getattr(ary, name),
+            self
+        )
+
+    def __getitem__(self, index):
+        return TracePair(self.dd,
+                         interior=self.interior[index],
+                         exterior=self.exterior[index])
```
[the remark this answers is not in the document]
thomasgibson: @inducer Could you check the wording to see if you're happy with these? https://github.com/inducer/grudge/pull/74/commits/cc158d41e4757428a51f9d0be016db8d2bd93494

grudge/geometry/metrics.py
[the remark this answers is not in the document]
thomasgibson: @inducer Addressed in this commit: https://github.com/inducer/grudge/pull/74/commits/2d74faaf0e1bbd619ebb7ede8d600aacb3d6468f. Could you let me know if you're happy with the approach?

grudge/trace_pair.py
```diff
@@ -0,0 +1,375 @@
...
+    exterior: ArrayContainer
+
+    def __init__(self, dd, *, interior, exterior):
+        object.__setattr__(self, "dd", dof_desc.as_dofdesc(dd))
+        object.__setattr__(self, "interior", interior)
+        object.__setattr__(self, "exterior", exterior)
+
+    def __getattr__(self, name):
+        """Return a :class:`TracePair` associated with the attributes
+        of the array containers defining :attr:`int` and :attr:`ext`.
```
inducer, line 112:
```suggestion
        """Return a new :class:`TracePair` resulting from executing attribute lookup
        with *name* on :attr:`int` and :attr:`ext`.
```

grudge/trace_pair.py
```diff
@@ -0,0 +1,375 @@
...
+        """
+        return TracePair(self.dd,
+                         interior=getattr(self.interior, name),
+                         exterior=getattr(self.exterior, name))
+
+    def __getitem__(self, index):
+        """Return a :class:`TracePair` associated with the subarrays
+        of :attr:`int` and :attr:`ext`, denoted by `index`.
```
inducer, line 120:
```suggestion
        """Return a new :class:`TracePair` resulting from executing subscripting with
        *index* on :attr:`int` and :attr:`ext`.
```

examples/wave/wave-op.py
```diff
@@ -157,13 +176,19 @@ def rhs(t, w):
...
-                  f"sol max: {op.nodal_max(dcoll, 'vol', fields[0])}")
-            vis.write_vtk_file("fld-wave-eager-%04d.vtu" % istep,
+            print(f"step: {istep} t: {t} "
+                  f"L2: {op.norm(dcoll, fields[0], 2)} "
+                  f"Linf: {op.norm(dcoll, fields[0], np.inf)} "
+                  f"sol max: {op.nodal_max(dcoll, 'vol', fields[0])} "
+                  f"sol min: {op.nodal_min(dcoll, 'vol', fields[0])}")
+            if write_output:
+                vis.write_vtk_file(
+                    "fld-wave-eager-%04d.vtu" % istep,
```
alexfikl, line 186:
```suggestion
                    f"fld-wave-eager-{istep:04d}.vtu",
```

grudge/discretization.py
```diff
@@ -21,45 +30,55 @@
...
 )
+
 import numpy as np  # noqa: F401
-from meshmode.array_context import ArrayContext
-from meshmode.discretization.connection import \
-    FACE_RESTR_INTERIOR, FACE_RESTR_ALL, make_face_restriction
-from meshmode.mesh import BTAG_PARTITION
 
-from warnings import warn
+from arraycontext.context import ArrayContext
```
alexfikl, line 41:
```suggestion
from arraycontext import ArrayContext
```

grudge/discretization.py
```diff
@@ -21,45 +30,55 @@
...
-"""
+from warnings import warn
 
 
 class DiscretizationCollection:
-    """
-    .. automethod :: __init__
+    """A collection of discretizations on various mesh entities
+    (volume, interior facets, boundaries) and associated element
+    groups.
```
alexfikl, line 56: Maybe mention somehow that all these discretizations are tied to the same underlying mesh?

grudge/discretization.py
```diff
@@ -21,45 +30,55 @@
...
     """
 
     def __init__(self, array_context, mesh, order=None,
             discr_tag_to_group_factory=None, mpi_communicator=None,
             # FIXME: `quad_tag_to_group_factory` is deprecated
             quad_tag_to_group_factory=None):
-        """
-        :param discr_tag_to_group_factory: A mapping from discretization tags
+        """Constructor for the :class:`DiscretizationCollection` object.
+
```
alexfikl, line 78:
```suggestion
        """
```

grudge/discretization.py
```diff
@@ -21,45 +30,55 @@
...
+        :arg actx: an :class:`~arraycontext.context.ArrayContext`.
+        :arg mesh: a :class:`~meshmode.mesh.Mesh` object.
```
alexfikl, line 80: Add some types instead of docs?

grudge/discretization.py
```diff
@@ -482,28 +533,49 @@ def _all_faces_volume_connection(self):
...
         return self._volume_discr.real_dtype
 
     @property
     def complex_dtype(self):
+        """Return the data type used for complex-valued arithmetic."""
         return self._volume_discr.complex_dtype
 
     @property
     def mesh(self):
+        """Return a :class:`meshmode.mesh.Mesh` over which the discretization
```
alexfikl, line 556:
```suggestion
        """Return the :class:`meshmode.mesh.Mesh` over which the discretization
```
?

grudge/discretization.py
```diff
@@ -482,28 +533,49 @@ def _all_faces_volume_connection(self):
...
+        collection is built.
+        """
         return self._volume_discr.mesh
 
     def empty(self, array_context: ArrayContext, dtype=None):
+        """Return an empty :class:`~meshmode.dof_array.DOFArray`.
```
alexfikl, line 562: Mention that it's on `DD_VOLUME`?

grudge/execution.py
```diff
@@ -20,6 +20,11 @@
 THE SOFTWARE.
 """
 
+
+from arraycontext.context import ArrayContext
+from arraycontext.loopy import make_loopy_program
+from arraycontext.container.traversal import thaw
```
alexfikl, line 26:
```suggestion
from arraycontext import ArrayContext, make_loopy_program, thaw
```

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,711 @@
...
+    return -form2.dot(inv_form1)
+
+
+def summed_curvature(actx: ArrayContext, dcoll: DiscretizationCollection,
+        dim=None, dd=None) -> DOFArray:
+    r"""Computes the sum of the principal curvatures:
+
+    .. math::
+
+        \kappa = \mathrm{Trace}(C(x))
```
alexfikl, line 684:
```suggestion
        \kappa = \operatorname{Trace}(C(x))
```

grudge/op.py
```diff
@@ -48,37 +84,74 @@
...
-def project(dcoll, src, tgt, vec):
+def project(dcoll: DiscretizationCollection, src, tgt, vec):
     """Project from one discretization to another, e.g. from the
     volume to the boundary, or from the base to the an overintegrated
     quadrature discretization.
 
-    :arg src: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one
-    :arg tgt: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`
+    :arg src: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
```
alexfikl, line 151: Might be worth introducing a type `DOFDescLike = Union[DOFDesc, Hashable]` or something like that? To avoid writing this everywhere.

alexfikl (approved):
I was curious what was going on here, so I gave it look! It's pretty big, so got lost towards the end, but left some comments along the way, mostly nitpicks with about 2-3 genuine questions.

Huge amount of work to clean this all up though! Nice!

grudge/trace_pair.py
thomasgibson: Thanks! Thanks! https://github.com/inducer/grudge/pull/74/commits/52c7318c777efebf2c559986f0f7b6a429898e89

grudge/trace_pair.py
thomasgibson: Thanks! https://github.com/inducer/grudge/pull/74/commits/52c7318c777efebf2c559986f0f7b6a429898e89

grudge/discretization.py
thomasgibson: Good suggestion! Added: https://github.com/inducer/grudge/pull/74/commits/52c7318c777efebf2c559986f0f7b6a429898e89

grudge/discretization.py
thomasgibson: https://github.com/inducer/grudge/commit/52c7318c777efebf2c559986f0f7b6a429898e89

grudge/discretization.py
thomasgibson: Done: https://github.com/inducer/grudge/commit/52c7318c777efebf2c559986f0f7b6a429898e89

grudge/discretization.py
thomasgibson: https://github.com/inducer/grudge/commit/52c7318c777efebf2c559986f0f7b6a429898e89

grudge/discretization.py
thomasgibson: Good call! https://github.com/inducer/grudge/commit/52c7318c777efebf2c559986f0f7b6a429898e89

grudge/geometry/metrics.py
thomasgibson: https://github.com/inducer/grudge/commit/52c7318c777efebf2c559986f0f7b6a429898e89

grudge/op.py
thomasgibson: I see what you mean. Though to be honest, I'd prefer to keep this explicit since, right now. It's technically not any hashable that works, only a very specific set of strings. In the future, we may want to consider being a bit more strict about what `dd` and others should be. For now, I'd prefer to leave this as is and just keep the docstring explicit.

grudge/models/advection.py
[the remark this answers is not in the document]
thomasgibson: Created an issue related to this so we can address it in a smaller future PR! https://github.com/inducer/grudge/issues/102

grudge/discretization.py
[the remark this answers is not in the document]
thomasgibson: Done here: https://github.com/inducer/grudge/pull/74/commits/86f08ababe29c3b91b4de88158045183ca35230d

inducer:
In 323b9fe, I've dropped the `dim` arguments throughout `geometry.metrics`. They were needed in the symbolic interface because the symbolics had no way of knowing the concrete dimension of the geometry. This code however can just *look* at the geometry to figure out dimensionality, so I didn't see much of a point in allowing it to override it. @alexfikl, objections?

grudge/geometry/metrics.py
```diff
@@ -0,0 +1,713 @@
...
+# }}}
+
+
+# {{{ Surface normal vectors
+
+def rel_mv_normal(
+        actx: ArrayContext, dcoll: DiscretizationCollection, dd=None) -> MultiVector:
+    r"""Computes surface normals at each nodal location as a
+    :class:`~pymbolic.geometric_algebra.MultiVector` relative to the 'volume'
+    discretization's surface pseudoscalar.
```
alexfikl, line 515:
```suggestion
    :class:`~pymbolic.geometric_algebra.MultiVector` relative to the
    surface pseudoscalar of the discretization described by *dd*.
```
? `'volume'` sounds a bit weird to me because this tends to compute normals on faces.

grudge/geometry/metrics.py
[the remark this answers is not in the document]
alexfikl: Left one little comment regarding the docs, but otherwise looks good to me!

alexfikl:
> In [323b9fe](https://github.com/inducer/grudge/commit/323b9fea1c0e7b1c8bfd857cef617f15d5590a2e), I've dropped the `dim` arguments throughout `geometry.metrics`. They were needed in the symbolic interface because the symbolics had no way of knowing the concrete dimension of the geometry. This code however can just _look_ at the geometry to figure out dimensionality, so I didn't see much of a point in allowing it to override it. @alexfikl, objections? (@thomasgibson, too, of course!)

Completely agreed! :rocket:

grudge/geometry/metrics.py
inducer: It does. #103 for the bigger issue.

alexfikl: I'm fine with just leaving this as is, since my suggestion isn't particularly clearer.

grudge/trace_pair.py
```diff
@@ -297,7 +310,8 @@ def finish(self):
                          exterior=swapped_remote_dof_array)
 
 
-def _cross_rank_trace_pairs_scalar_field(dcoll, vec, tag=None):
+def _cross_rank_trace_pairs_scalar_field(
+        dcoll: DiscretizationCollection, vec, tag=None) -> list:
```
inducer, line 314: @thomasgibson FYI these could be `List[TracePair]`. (I won't let this hold up the merge.)

inducer:
4bd6d3b0361bb192fafd21fb75afbea1347c6d95 is an important thing that I'm glad I caught, and I think the idiom I ended up using works reasonably well ("freeze before memoizing, but return a thawed value").

It's important to memoize frozen data because
* for PyOpenCL arrays, they might use incorrect synchronization (with a different queue), or (hopefully) complain of queue mismatches, and
* for lazy arrays, you might end up redoing the computation every time.

cc @thomasgibson

inducer:
Whew! What a PR! :) Thanks for all your work on this, and for seeing it through to the end! I'm super happy with how it turned out. It'll land as soon as CI passes. :tada:

inducer:
It's in! :D

PR #75, opened 2021-04-25
Untangle `DOFDesc` and related tags from the symbolic infrastructure

thomasgibson:
Some refactoring/untangling of the `DOFDesc` and related objects. I'm also a firm believer that little documentation is better than no documentation, even for extremely simple classes. So I've taken this opportunity to add some docstrings to a few of the objects commonly used in the DOF description interface. Please feel free to leave comments to suggest any improvements to the docs as well.


inducer (approved):
LGTM, thanks!

PR #76, opened 2021-04-26
Do not cache forward mapping derivative in discretization

inducer:


inducer:
Turns out that removing the forward metric derivatives exposed a "compiler" bug that (I think) was previously reported: https://gitlab.tiker.net/inducer/grudge/-/merge_requests/67. cc @alexfikl

alexfikl:
> Turns out that removing the forward metric derivatives exposed a "compiler" bug that (I think) was previously reported: [gitlab.tiker.net/inducer/grudge/-/merge_requests/67](https://gitlab.tiker.net/inducer/grudge/-/merge_requests/67). cc @alexfikl

Ha! Haven't seen that in a while! I'll try and revive that PR and see if it still crashes.

alexfikl:
I'm a bit surprised this doesn't have any noticeable performance impact. Wouldn't these things get recomputed a lot for normals, area elements, etc?

inducer:
> I'm a bit surprised this doesn't have any noticeable performance impact. Wouldn't these things get recomputed a lot for normals, area elements, etc?

[Normals](https://github.com/inducer/grudge/blob/b54af213d7a063afc9cce179e9e4a208971bfe6d/grudge/symbolic/primitives.py#L610-L612) and [area elements](https://github.com/inducer/grudge/blob/b54af213d7a063afc9cce179e9e4a208971bfe6d/grudge/symbolic/primitives.py#L589-L593) are cached separately.

alexfikl:
> [Normals](https://github.com/inducer/grudge/blob/b54af213d7a063afc9cce179e9e4a208971bfe6d/grudge/symbolic/primitives.py#L610-L612) and [area elements](https://github.com/inducer/grudge/blob/b54af213d7a063afc9cce179e9e4a208971bfe6d/grudge/symbolic/primitives.py#L589-L593) are cached separately.

Ah, so even if they get computed once for the normal and once for the area elements, it's still a one time cost that doesn't show up much. Makes sense, thanks!

PR #77, opened 2021-04-26
Add hooks for constructing connections to and from modal representations

thomasgibson:
This brings in the recent modal connection work in meshmode to grudge. Can be used to simplify the modal indicator used in https://github.com/illinois-ceesd/mirgecom/pull/213 and modal filtering in https://github.com/illinois-ceesd/mirgecom/pull/92

@MTCam @lukeolson @inducer


grudge/dof_desc.py
```diff
@@ -58,6 +60,13 @@ class DTAG_VOLUME_ALL:  # noqa: N801
     """
 
 
+class DTAG_MODAL:  # noqa: N801
+    """
+    A tag denoting modal (rather than nodal)
+    degrees of freedom on cell volumes.
+    """
```
inducer, line 67: This is not the right place for this. `DTAG` stands for "domain tag" (as you can tell from the surroundings), i.e. which geometry this is on. `QTAG` ("quadrature tag") describes how something is discretized. It's obviously possible to use modal discretizations on boundaries and volumes, so connecting these two concerns does not make sense.

`QTAG_MODAL` would be better, however "quadrature" is a misnomer. I'm open to renaming `QTAG`.

thomasgibson: I vote for `QTAG_MODAL` for now.

thomasgibson: To be fair, `DTAG_SCALAR` doesn't seem to make sense as a "domain tag" either. But I'll leave that for another day.

inducer: > To be fair, `DTAG_SCALAR` doesn't seem to make sense as a "domain tag" either. But I'll leave that for another day.

Why doesn't it? The domain of the scalar is a set (any set) with a single element.

thomasgibson: Fixed in https://github.com/inducer/grudge/pull/77/commits/e9f6da0d85da1802a2b3f9b43ac32c5173ef65d9

thomasgibson: > Why doesn't it? The domain of the scalar is a set (any set) with a single element.

I think I was just incorrectly trying to pin too much "fem" type concepts to the tag when I initially looked at this. Disregard what I said. It's my misunderstanding.

inducer: #79

grudge/discretization.py
```diff
@@ -239,10 +261,10 @@ def connection_from_dds(self, from_dd, to_dd):
 
         # {{{ generic to-quad
 
-        if (
-                from_dd.domain_tag == to_dd.domain_tag
-                and from_dd.quadrature_tag is QTAG_NONE
-                and to_dd.quadrature_tag is not QTAG_NONE):
+        if (from_dd.domain_tag == to_dd.domain_tag
```
inducer, line 264:
```suggestion
        # QTAG_MODAL is handled above
        if (from_dd.domain_tag == to_dd.domain_tag
```

grudge/dof_desc.py
```diff
@@ -36,12 +36,14 @@
...
 .. autoclass:: DTAG_BOUNDARY
 .. autoclass:: QTAG_NONE
+.. autoclass:: QTAG_MODAL
 
 .. autoclass:: DOFDesc
 .. autofunction:: as_dofdesc
 
 .. data:: DD_SCALAR
 .. data:: DD_VOLUME
+.. data:: DD_MODAL
```
inducer, line 46:
```suggestion
.. data:: DD_VOLUME_MODAL
```
? I'm also OK with `DD_VOL_MODAL`.

inducer (commented):
Thanks! Two minor things, otherwise LGTM.

inducer:
Thanks!

PR #78, opened 2021-04-27
Add tests for some code generation bugs

alexfikl:
Updated from https://gitlab.tiker.net/inducer/grudge/-/merge_requests/67

PR #80, opened 2021-04-27
Add tests for some code generation bugs

inducer:
It's #78, but revived.

cc @alexfikl


inducer:
Here it won't let me update the branch because I created the PR.

PR #81, opened 2021-04-27
Add tests for some code generation bugs

inducer:
It's #78 and #80, but revived in such a way that I can actually merge them. Grrr, Github UI.

cc @alexfikl


alexfikl:
I don't think I can push to this, but it should be mostly done anyway. Was there anything else you wanted changed?

inducer:
Nope, just forgot to click automerge. Thanks again!

PR #83, opened 2021-04-27
Introduce discretization tags in place of quadrature tags

### LINKED ISSUE #79: QTAG_XXX is misnamed
With the introduction of `QTAG_MODAL` in #77, it's apparent that the `QTAG_` prefixes (for "quadrature tag") are misnamed. "Modal" just isn't a form of quadrature, unless you twist your head in a very special way. "Discretization tag" would make sense, but the `DTAG_` is taken ("domain tag"). `DISCR_TAG_XXX`?

cc @thomasgibson @majosm @alexfikl

thomasgibson:
Drops `QTAG_` in favor of `DISCR_TAG_`

cc @majosm @alexfikl @inducer 

Closes #79.

### DIFF
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -22,7 +22,7 @@ THE SOFTWARE.
 
 from pytools import memoize_method
 from grudge.dof_desc import (
-    DISCR_TAG_BASE, DISCR_TAG_QUAD, DISCR_TAG_MODAL,
+    DISCR_TAG_BASE, DISCR_TAG_MODAL,
     DTAG_BOUNDARY, DOFDesc, as_dofdesc,
     # importing so we can yell at the user for using it
     QTAG_NONE
@@ -171,7 +171,7 @@ class DiscretizationCollection:
         return boundary_connections
 
     def get_distributed_boundary_swap_connection(self, dd):
-        if not dd.discretization_tag in (DISCR_TAG_BASE, DISCR_TAG_MODAL):
+        if dd.discretization_tag not in (DISCR_TAG_BASE, DISCR_TAG_MODAL):
             # FIXME
             raise NotImplementedError("Distributed communication with quadrature")
 
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -142,6 +142,16 @@ class DiscretizationCollection:
 
         self.mpi_communicator = mpi_communicator
 
+    @property
+    def quad_tag_to_group_factory(self):
+        warn("`DiscretizationCollection.quad_tag_to_group_factory` "
+             "is deprecated and will go away in 2022. Use "
+             "`DiscretizationCollection.discr_tag_to_group_factory` "
+             "instead.",
+             DeprecationWarning, stacklevel=2)
+
+        return self.discr_tag_to_group_factory
+
     def get_management_rank_index(self):
         return 0
 
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -90,14 +90,12 @@ class DiscretizationCollection:
                         PolynomialWarpAndBlendGroupFactory(order=order)
 
         # FIXME: QTAG_NONE hunting
-        if QTAG_NONE in quad_tag_to_group_factory.keys():
+        if QTAG_NONE in quad_tag_to_group_factory:
             warn("`DOFDesc.QTAG_NONE` is deprecated and will be dropped "
                  "in version 2022.x. Use `DOFDesc.DISCR_TAG_BASE` instead.",
                  DeprecationWarning, stacklevel=2)
-            # Is this too aggressive/invasive?
             quad_tag_to_group_factory[DISCR_TAG_BASE] = \
                 quad_tag_to_group_factory[QTAG_NONE]
-            del quad_tag_to_group_factory[QTAG_NONE]
 
         # Modal discr should always comes from the base discretization
         quad_tag_to_group_factory[DISCR_TAG_MODAL] = \
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -186,7 +186,7 @@ class DiscretizationCollection:
 
         if dd.is_volume():
             if discr_tag is not DISCR_TAG_BASE:
-                return self._quad_volume_discr(discr_tag)
+                return self._discr_tag_volume_discr(discr_tag)
             return self._volume_discr
 
         if discr_tag is not DISCR_TAG_BASE:
@@ -306,7 +306,7 @@ class DiscretizationCollection:
             elif to_dd.is_volume():
                 from meshmode.discretization.connection import \
                         make_same_mesh_connection
-                to_discr = self._quad_volume_discr(to_discr_tag)
+                to_discr = self._discr_tag_volume_discr(to_discr_tag)
                 from_discr = self._volume_discr
                 return make_same_mesh_connection(self._setup_actx, to_discr,
                             from_discr)
@@ -336,7 +336,7 @@ class DiscretizationCollection:
         return self.quad_tag_to_group_factory[discretization_tag]
 
     @memoize_method
-    def _quad_volume_discr(self, discretization_tag):
+    def _discr_tag_volume_discr(self, discretization_tag):
         from meshmode.discretization import Discretization
 
         return Discretization(
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -160,9 +160,12 @@ class DiscretizationCollection:
         return boundary_connections
 
     def get_distributed_boundary_swap_connection(self, dd):
-        if dd.discretization_tag not in (DISCR_TAG_BASE, DISCR_TAG_MODAL):
+        if dd.discretization_tag is not DISCR_TAG_BASE:
             # FIXME
-            raise NotImplementedError("Distributed communication with quadrature")
+            raise NotImplementedError(
+                "Distributed communication with discretization tag "
+                f"{dd.discretization_tag} is not implemented."
+            )
 
         assert isinstance(dd.domain_tag, DTAG_BOUNDARY)
         assert isinstance(dd.domain_tag.tag, BTAG_PARTITION)
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -23,9 +23,7 @@ THE SOFTWARE.
 from pytools import memoize_method
 from grudge.dof_desc import (
     DISCR_TAG_BASE, DISCR_TAG_MODAL,
-    DTAG_BOUNDARY, DOFDesc, as_dofdesc,
-    # importing so we can yell at the user for using it
-    QTAG_NONE
+    DTAG_BOUNDARY, DOFDesc, as_dofdesc
 )
 import numpy as np  # noqa: F401
 from meshmode.array_context import ArrayContext
@@ -88,14 +86,6 @@ class DiscretizationCollection:
 
                 quad_tag_to_group_factory[DISCR_TAG_BASE] = \
                         PolynomialWarpAndBlendGroupFactory(order=order)
-
-        # FIXME: QTAG_NONE hunting
-        if QTAG_NONE in quad_tag_to_group_factory:
-            warn("`DOFDesc.QTAG_NONE` is deprecated and will be dropped "
-                 "in version 2022.x. Use `DOFDesc.DISCR_TAG_BASE` instead.",
-                 DeprecationWarning, stacklevel=2)
-            quad_tag_to_group_factory[DISCR_TAG_BASE] = \
-                quad_tag_to_group_factory[QTAG_NONE]
 
         # Modal discr should always comes from the base discretization
         quad_tag_to_group_factory[DISCR_TAG_MODAL] = \
@@ -183,13 +173,6 @@ class DiscretizationCollection:
     def discr_from_dd(self, dd):
         dd = as_dofdesc(dd)
 
-        # FIXME: QTAG_NONE hunting
-        if dd.discretization_tag is QTAG_NONE:
-            warn("`DOFDesc.QTAG_NONE` is deprecated and will be dropped "
-                 "in version 2022.x. Use `DOFDesc.DISCR_TAG_BASE` instead.",
-                 DeprecationWarning, stacklevel=2)
-            dd = dd.with_discr_tag(DISCR_TAG_BASE)
-
         discr_tag = dd.discretization_tag
 
         if discr_tag is DISCR_TAG_MODAL:
@@ -224,20 +207,6 @@ class DiscretizationCollection:
     def connection_from_dds(self, from_dd, to_dd):
         from_dd = as_dofdesc(from_dd)
         to_dd = as_dofdesc(to_dd)
-
-        # FIXME: QTAG_NONE hunting
-        if to_dd.discretization_tag is QTAG_NONE:
-            warn("`DOFDesc.QTAG_NONE` is deprecated and will be dropped "
-                 "in version 2022.x. Use `DOFDesc.DISCR_TAG_BASE` instead.",
-                 DeprecationWarning, stacklevel=2)
-            to_dd = to_dd.with_discr_tag(DISCR_TAG_BASE)
-
-        # FIXME: QTAG_NONE hunting
-        if from_dd.discretization_tag is QTAG_NONE:
-            warn("`DOFDesc.QTAG_NONE` is deprecated and will be dropped "
-                 "in version 2022.x. Use `DOFDesc.DISCR_TAG_BASE` instead.",
-                 DeprecationWarning, stacklevel=2)
-            from_dd = from_dd.with_discr_tag(DISCR_TAG_BASE)
 
         to_discr_tag = to_dd.discretization_tag
         from_discr_tag = from_dd.discretization_tag
@@ -345,13 +314,6 @@ class DiscretizationCollection:
         """
         OK to override in user code to control mode/node choice.
         """
-        # FIXME: QTAG_NONE hunting
-        if discretization_tag is QTAG_NONE:
-            warn("`DOFDesc.QTAG_NONE` is deprecated and will be dropped "
-                 "in version 2022.x. Use `DOFDesc.DISCR_TAG_BASE` instead.",
-                 DeprecationWarning, stacklevel=2)
-            discretization_tag = DISCR_TAG_BASE
-
         if discretization_tag is None:
             discretization_tag = DISCR_TAG_BASE
 
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -41,6 +41,8 @@ __doc__ = """
 
 class DiscretizationCollection:
     """
+    .. automethod :: __init__
+
     .. automethod :: discr_from_dd
     .. automethod :: connection_from_dds
 
@@ -53,47 +55,69 @@ class DiscretizationCollection:
     """
 
     def __init__(self, array_context, mesh, order=None,
-            quad_tag_to_group_factory=None, mpi_communicator=None):
+            discr_tag_to_group_factory=None, mpi_communicator=None,
+            # FIXME: `quad_tag_to_group_factory` is deprecated
+            quad_tag_to_group_factory=None):
         """
-        :param quad_tag_to_group_factory: A mapping from quadrature tags (typically
-            strings--but may be any hashable/comparable object) to a
+        :param discr_tag_to_group_factory: A mapping from discretization tags
+            (typically one of: :class:`grudge.dof_desc.DISCR_TAG_BASE`,
+            :class:`grudge.dof_desc.DISCR_TAG_MODAL`, or
+            :class:`grudge.dof_desc.DISCR_TAG_QUAD`) to a
             :class:`~meshmode.discretization.poly_element.ElementGroupFactory`
-            indicating with which quadrature discretization the operations are
+            indicating with which type of discretization the operations are
             to be carried out, or *None* to indicate that operations with this
-            quadrature tag should be carried out with the standard volume
+            discretization tag should be carried out with the standard volume
             discretization.
         """
+
+        if (quad_tag_to_group_factory is not None
+                and discr_tag_to_group_factory is not None):
+            raise ValueError(
+                "Both `quad_tag_to_group_factory` and `discr_tag_to_group_factory` "
+                "are specified. Use `discr_tag_to_group_factory` instead."
+            )
+
+        # FIXME: `quad_tag_to_group_factory` is deprecated
+        if (quad_tag_to_group_factory is not None
+                and discr_tag_to_group_factory is None):
+            warn("`quad_tag_to_group_factory` is a deprecated kwarg and will "
+                 "be dropped in version 2022.x. Use `discr_tag_to_group_factory` "
+                 "instead.",
+                 DeprecationWarning, stacklevel=2)
+            discr_tag_to_group_factory = quad_tag_to_group_factory
 
         self._setup_actx = array_context
 
         from meshmode.discretization.poly_element import \
                 PolynomialWarpAndBlendGroupFactory
 
-        if quad_tag_to_group_factory is None:
+        if discr_tag_to_group_factory is None:
             if order is None:
-                raise TypeError("one of 'order' and "
-                        "'quad_tag_to_group_factory' must be given")
+                raise TypeError(
+                    "one of 'order' and 'discr_tag_to_group_factory' must be given"
+                )
 
-            quad_tag_to_group_factory = {
+            discr_tag_to_group_factory = {
                     DISCR_TAG_BASE: PolynomialWarpAndBlendGroupFactory(order=order)}
         else:
             if order is not None:
-                quad_tag_to_group_factory = quad_tag_to_group_factory.copy()
-                if DISCR_TAG_BASE in quad_tag_to_group_factory:
-                    raise ValueError("if 'order' is given, "
-                            "'quad_tag_to_group_factory' must not have a "
-                            "key of DISCR_TAG_BASE")
+                discr_tag_to_group_factory = discr_tag_to_group_factory.copy()
+                if DISCR_TAG_BASE in discr_tag_to_group_factory:
+                    raise ValueError(
+                        "if 'order' is given, 'discr_tag_to_group_factory' must "
+                        "not have a key of DISCR_TAG_BASE"
+                    )
 
-                quad_tag_to_group_factory[DISCR_TAG_BASE] = \
+                discr_tag_to_group_factory[DISCR_TAG_BASE] = \
                         PolynomialWarpAndBlendGroupFactory(order=order)
 
         # Modal discr should always comes from the base discretization
-        quad_tag_to_group_factory[DISCR_TAG_MODAL] = \
+        discr_tag_to_group_factory[DISCR_TAG_MODAL] = \
             _generate_modal_group_factory(
-                quad_tag_to_group_factory[DISCR_TAG_BASE]
+                discr_tag_to_group_factory[DISCR_TAG_BASE]
             )
 
-        self.quad_tag_to_group_factory = quad_tag_to_group_factory
+        self.discr_tag_to_group_factory = discr_tag_to_group_factory
 
         from meshmode.discretization import Discretization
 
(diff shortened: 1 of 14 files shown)


grudge/discretization.py
```diff
@@ -71,30 +75,41 @@ def __init__(self, array_context, mesh, order=None,
...
-                            "key of QTAG_NONE")
+                            "key of DISCR_TAG_BASE")
 
-                quad_tag_to_group_factory[QTAG_NONE] = \
+                quad_tag_to_group_factory[DISCR_TAG_BASE] = \
                         PolynomialWarpAndBlendGroupFactory(order=order)
 
+        # FIXME: Get rid of QTAG_NONE in downstream grudge applications
+        # Is this too aggressive/invasive?
+        if QTAG_NONE in quad_tag_to_group_factory.keys():
```
alexfikl, line 92:
```suggestion
        if QTAG_NONE in quad_tag_to_group_factory:
```

grudge/discretization.py
```diff
@@ -71,30 +75,41 @@ def __init__(self, array_context, mesh, order=None,
...
+            from warnings import warn
+            warn("`DOFDesc.QTAG_NONE` is deprecated and will be dropped "
+                 "in version 2022.x. Use `DOFDesc.DISCR_TAG_BASE` instead.",
+                 DeprecationWarning, stacklevel=2)
+            quad_tag_to_group_factory[DISCR_TAG_BASE] = \
+                quad_tag_to_group_factory[QTAG_NONE]
+            del quad_tag_to_group_factory[QTAG_NONE]
```
alexfikl, line 99: If the user gives the `quad_tag_to_group_factory` dict, they may expect `QTAG_NONE` to still be in there, so I wouldn't delete it.

grudge/dof_desc.py
```diff
@@ -96,29 +98,44 @@ def __repr__(self):
...
     basic discretization grid. This tag is used
-    to distinguish the base discretization (`QTAG_NONE`)
-    from quadrature (e.g. overintegration) grids.
+    to distinguish the base discretization from quadrature
+    (e.g. overintegration) or modal (`DISCR_TAG_MODAL`)
+    discretizations.
+    """
+
+
+# FIXME: This needs to go away
```
alexfikl, line 110:
```suggestion
# FIXME: This is deprecated and will be removed
```
?

grudge/dof_desc.py
```diff
@@ -96,29 +98,44 @@ def __repr__(self):
...
+class DISCR_TAG_MODAL:  # noqa: N801
+    """A discretization tag indicating the use of a
     basic discretization grid with modal degrees of
     freedom. This tag is used to distinguish the
-    modal discretization (`QTAG_MODAL`) from
-    the base (nodal) discretization (e.g. `QTAG_NONE`)
-    or discretizations on quadrature grids.
+    modal discretization from the base (nodal)
+    discretization (e.g. `DISCR_TAG_BASE`) or
+    discretizations on quadrature grids (`DISCR_TAG_QUAD`).
```
alexfikl, line 130:
```suggestion
    discretization (e.g. :class:`DISCR_TAG_BASE`) or
    discretizations on quadrature grids (:class:`DISCR_TAG_QUAD`).
```

grudge/dof_desc.py
```diff
@@ -130,15 +147,15 @@ class DOFDesc:
...
+    .. automethod:: with_discr_tag
     .. automethod:: with_dtag
 
     .. automethod:: __eq__
     .. automethod:: __ne__
     .. automethod:: __hash__
     """
 
-    def __init__(self, domain_tag, quadrature_tag=None):
+    def __init__(self, domain_tag, discretization_tag=None):
```
alexfikl, line 158:
```suggestion
    def __init__(self, domain_tag, discretization_tag=None, quadrature_tag=None):
```
Maybe keep it for backwards compatibility, in case someone called with the keyword? This would also need some more work in the constructor to warn.

alexfikl (approved):
Looks good to me, barring some small nitpicks!

grudge/dof_desc.py
thomasgibson: Done in https://github.com/inducer/grudge/pull/83/commits/4ed7e6f15401b6cb5da85c84f45ec267dbc2db1f!

grudge/dof_desc.py
thomasgibson: Done in https://github.com/inducer/grudge/pull/83/commits/4ed7e6f15401b6cb5da85c84f45ec267dbc2db1f

grudge/dof_desc.py
thomasgibson: Done in https://github.com/inducer/grudge/pull/83/commits/4ed7e6f15401b6cb5da85c84f45ec267dbc2db1f

grudge/discretization.py
thomasgibson: Done in https://github.com/inducer/grudge/pull/83/commits/826f66fc3b3b6e7cc67dc16e40946bbc23361e9e

grudge/discretization.py
thomasgibson: Done in https://github.com/inducer/grudge/pull/83/commits/826f66fc3b3b6e7cc67dc16e40946bbc23361e9e

thomasgibson:
There were several places I needed to add some checking for `QTAG_NONE`. I've added comments `# FIXME: QTAG_NONE hunting` in relevant places.

thomasgibson:
Oh boy --- debugging downstream mirgecom now :sob:

thomasgibson:
OK! _Now_ it should be good.

thomasgibson:
@alexfikl @inducer --- I think this needs another glance due to all the `QTAG_` catching in order for things to not break with the tag changes. Just to make sure you are all happy with how it's done!

alexfikl:
> @alexfikl @inducer --- I think this needs another glance due to all the `QTAG_` catching in order for things to not break with the tag changes. Just to make sure you are all happy with how it's done!

Could we just set `QTAG_NONE = DISCR_TAG_BASE` and `QTAG_MODAL = DISCR_TAG_MODAL` and avoid all of that? 

The only downside I can think of is that we won't get a nice deprecation warning, but python 3.7 has module level `__getattr__` with [PEP 562](https://www.python.org/dev/peps/pep-0562/) which could be used for that too. What do you think?

thomasgibson:
> > @alexfikl @inducer --- I think this needs another glance due to all the `QTAG_` catching in order for things to not break with the tag changes. Just to make sure you are all happy with how it's done!
> 
> Could we just set `QTAG_NONE = DISCR_TAG_BASE` and `QTAG_MODAL = DISCR_TAG_MODAL` and avoid all of that?
> 
> The only downside I can think of is that we won't get a nice deprecation warning, but python 3.7 has module level `__getattr__` with [PEP 562](https://www.python.org/dev/peps/pep-0562/) which could be used for that too. What do you think?

Oh this seems much nicer, I'll take a look at this and try it out

thomasgibson:
@alexfikl I really want to use the PEP 562 approach, but I was unable to get it to do anything... I'm not sure why. I've added this to the `dof_desc.py`:
```python
_deprecated_names = {"QTAG_NONE": "DISCR_TAG_BASE",
                     "QTAG_MODAL": "DISCR_TAG_MODAL"}


def __getattr__(name):
    if name in _deprecated_names:
        warn(f"{name} is a deprecated kwarg and will be dropped "
             f"in version 2022.x. Use {_deprecated_names[name]} instead",
             DeprecationWarning, stacklevel=2)
        return globals()[f"{name}"]
    raise AttributeError(f"module {__name__} has no attribute {name}")
```

But each time I try the following:
```python
from grudge.dof_desc import QTAG_NONE
```
I don't get the deprecation warning I was hoping for

alexfikl:
My bad, the `__getattr__` only gets called if the attribute is not defined, so you can't have `QTAG_NONE = DISCR_TAG_BASE` in the module. This patch makes it work for me
```patch
diff --git a/grudge/dof_desc.py b/grudge/dof_desc.py
index 13db8e8..3e0ddbc 100644
--- a/grudge/dof_desc.py
+++ b/grudge/dof_desc.py
@@ -49,6 +49,23 @@ __doc__ = """
 .. data:: DD_VOLUME_MODAL
 """
 
+# {{{ deprecation warnings
+
+_deprecated_names = {"QTAG_NONE": "DISCR_TAG_BASE",
+                     "QTAG_MODAL": "DISCR_TAG_MODAL"}
+
+
+def __getattr__(name):
+    if name in _deprecated_names:
+        warn(f"{name} is a deprecated kwarg and will be dropped "
+             f"in version 2022.x. Use {_deprecated_names[name]} instead",
+             DeprecationWarning, stacklevel=2)
+        return globals()[f"{_deprecated_names[name]}"]
+
+    raise AttributeError(f"module {__name__} has no attribute {name}")
+
+# }}}
+
 
 # {{{ DOF description
 
@@ -303,14 +320,4 @@ def as_dofdesc(dd):
 
 # }}}
 
-
-# {{{ Deprecated tags
-
-# FIXME: These are deprecated and will be removed
-QTAG_NONE = DISCR_TAG_BASE
-QTAG_MODAL = DISCR_TAG_MODAL
-
-# }}}
-
-
 # vim: foldmethod=marker
```

thomasgibson:
> My bad, the `__getattr__` only gets called if the attribute is not defined, so you can't have `QTAG_NONE = DISCR_TAG_BASE` in the module. This patch makes it work for me
> 
> ```diff
> diff --git a/grudge/dof_desc.py b/grudge/dof_desc.py
> index 13db8e8..db0be2b 100644
> --- a/grudge/dof_desc.py
> +++ b/grudge/dof_desc.py
> @@ -25,6 +25,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
>  THE SOFTWARE.
>  """
>  
> +import sys
> +
>  from meshmode.discretization.connection import \
>      FACE_RESTR_INTERIOR, FACE_RESTR_ALL
>  from meshmode.mesh import \
> @@ -49,6 +51,23 @@ __doc__ = """
>  .. data:: DD_VOLUME_MODAL
>  """
>  
> +# {{{ deprecation warnings
> +
> +_deprecated_names = {"QTAG_NONE": "DISCR_TAG_BASE",
> +                     "QTAG_MODAL": "DISCR_TAG_MODAL"}
> +
> +
> +def __getattr__(name):
> +    if name in _deprecated_names:
> +        warn(f"{name} is deprecated and will be dropped "
> +             f"in version 2022.x. Use {_deprecated_names[name]} instead.",
> +             DeprecationWarning, stacklevel=2)
> +        return globals()[f"{_deprecated_names[name]}"]
> +
> +    raise AttributeError(f"module {__name__} has no attribute {name}")
> +
> +# }}}
> +
>  
>  # {{{ DOF description
>  
> @@ -307,10 +326,10 @@ def as_dofdesc(dd):
>  # {{{ Deprecated tags
>  
>  # FIXME: These are deprecated and will be removed
> -QTAG_NONE = DISCR_TAG_BASE
> -QTAG_MODAL = DISCR_TAG_MODAL
> +if sys.version_info < (3, 7):
> +    QTAG_NONE = DISCR_TAG_BASE
> +    QTAG_MODAL = DISCR_TAG_MODAL
>  
>  # }}}
>  
> -
>  # vim: foldmethod=marker
> ```

Works for me as well! Thanks!

grudge/dof_desc.py
```diff
@@ -241,20 +286,45 @@ def fmt(s):
...
+    if name in _deprecated_names:
+        warn(f"{name} is deprecated and will be dropped "
+             f"in version 2022.x. Use {_deprecated_names[name]} instead.",
+             DeprecationWarning, stacklevel=2)
+        return globals()[f"{_deprecated_names[name]}"]
+
+    raise AttributeError(f"module {__name__} has no attribute {name}")
+
+
+import sys
```
alexfikl, line 323: Move the import to the top?

grudge/symbolic/operators.py
```diff
@@ -103,8 +104,11 @@ class Operator(pymbolic.primitives.Expression):
 
     def __init__(self, dd_in, dd_out):
         import grudge.dof_desc as dof_desc
-        self.dd_in = dof_desc.as_dofdesc(dd_in)
-        self.dd_out = dof_desc.as_dofdesc(dd_out)
+        dd_in = dof_desc.as_dofdesc(dd_in)
+        dd_out = dof_desc.as_dofdesc(dd_out)
+
+        self.dd_in = dd_in
+        self.dd_out = dd_out
```
alexfikl, line 111:
```suggestion
        self.dd_in = dof_desc.as_dofdesc(dd_in)
        self.dd_out = dof_desc.as_dofdesc(dd_out)
```
? Unless I misunderstand why they were separated.

alexfikl (approved):
Just a couple of nitpicks from me, otherwise it looks good!

grudge/symbolic/operators.py
thomasgibson: Whoops! This was leftover from the previous `QTAG_NONE` hunting spree. I'll fix this!

grudge/dof_desc.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/f0b8bdf2bf05f13c019530025a3039afc045f6a2

grudge/symbolic/operators.py
thomasgibson: Reverted this (and similar no-op changes) in: https://github.com/inducer/grudge/pull/83/commits/360ed3a7853c5cf2ab24bce93d40d785f7933be6

grudge/discretization.py
```diff
@@ -156,7 +160,7 @@ def _set_up_distributed_communication(self, mpi_communicator, array_context):
         return boundary_connections
 
     def get_distributed_boundary_swap_connection(self, dd):
-        if dd.quadrature_tag != QTAG_NONE:
+        if dd.discretization_tag not in (DISCR_TAG_BASE, DISCR_TAG_MODAL):
             # FIXME
             raise NotImplementedError("Distributed communication with quadrature")
```
inducer, line 165: If you ask `connection_from_dds` for a connection modal volume -> nodal boundary, does that work? Also, can you ask for a "face->volume mass matrix" onto the modal discretization? (If the answer to either of these is no, then this shouldn't suggest that modal is allowed here.)

grudge/dof_desc.py
```diff
@@ -180,14 +197,35 @@ def __init__(self, domain_tag, quadrature_tag=None):
...
 
-        if domain_tag is DTAG_SCALAR and quadrature_tag is not None:
-            raise ValueError("cannot have nontrivial quadrature tag on scalar")
+        # Prints warning if a user passes both a quadrature_tag and
+        # a discretization_tag
+        if (quadrature_tag is not None and discretization_tag is not None):
+            warn("`quadrature_tag` and `discretization_tag` are specified. "
+                 "`quadrature_tag` will be ignored. Use `discretization_tag` "
+                 "instead.",
+                 DeprecationWarning, stacklevel=2)
```
inducer, line 206: This should be an error IMO.

grudge/dof_desc.py
```diff
@@ -96,29 +100,39 @@ def __repr__(self):
...
 
-class QTAG_MODAL:  # noqa: N801
-    """A quadrature tag indicating the use of a
+class DISCR_TAG_QUAD:  # noqa: N801
+    """A discretization tag indicating the use of a
+    quadrature discretization grid. This tag is used
+    to distinguish the quadrature discretization
+    (e.g. overintegration) from modal (:class:`DISCR_TAG_MODAL`)
+    or base (:class:`DISCR_TAG_BASE`) discretizations.
+    """
```
inducer, line 118: Do you want to include a blurb about subclassing to accomplish more than one quadrature tag?

grudge/dof_desc.py
```diff
@@ -154,11 +170,12 @@ def __init__(self, domain_tag, quadrature_tag=None):
             :class:`~meshmode.mesh.BTAG_PARTITION`,
             or *None* to indicate that the geometry is not yet known.
 
-        :arg quadrature_tag:
-            *None* to indicate that the quadrature grid is not known, or
-            :class:`QTAG_NONE` to indicate the use of the basic discretization
-            grid, or a string to indicate the use of the thus-tagged quadratue
-            grid.
+        :arg discretization_tag:
+            *None* to indicate that the discretization grid is not known,
```
inducer, line 174: > not known

What would that mean? (FWIW, I think `None`/"not specified" should mean "base" as the logic below says.)

grudge/dof_desc.py
```diff
@@ -154,11 +170,12 @@ def __init__(self, domain_tag, quadrature_tag=None):
...
+            :class:`DISCR_TAG_BASE` to indicate the use of the basic
+            discretization grid, :class:`DISCR_TAG_MODAL` to indicate a
+            modal discretization, or :class:`DISCR_TAG_QUAD` or a string
+            to indicate the use of the thus-tagged quadrature grid.
```
inducer, line 178: I'd say we deprecate (and don't document) string discretization tags.

grudge/dof_desc.py
```diff
@@ -241,20 +287,43 @@ def fmt(s):
...
+             f"in version 2022.x. Use {_deprecated_names[name]} instead.",
+             DeprecationWarning, stacklevel=2)
+        return globals()[f"{_deprecated_names[name]}"]
+
+    raise AttributeError(f"module {__name__} has no attribute {name}")
+
+
+if sys.version_info < (3, 7):
+    QTAG_NONE = DISCR_TAG_BASE
+    QTAG_MODAL = DISCR_TAG_MODAL
```
inducer, line 326: This could just be a loop over `_deprecated_names` that updates `globals()`, to avoid two (possibly inconsistent) sources of truth.

inducer (commented):
Thanks! I'm excited about this. A few (hopefully simple) things below.

inducer:

grudge/dof_desc.py
thomasgibson: Ah I just realized something. So all these tags (`DISCR_TAG_`, `DTAG_`, etc) are never used as instantiated objects. The only exception is `DTAG_BOUNDARY` I think. Do you think `DISCR_TAG_QUAD` should follow a similar behavior? That way we can check `isinstance(tag, DISCR_TAG_QUAD)` if we ever want to check if a quadrature tag is in use (and similarly in other places).

What do you think @inducer ?

alexfikl: You could use [issubclass](https://docs.python.org/3/library/functions.html#issubclass) to just check classes, not instances?

inducer: I think instances for `DTAG_QUAD` make sense. It doesn't preclude subclassing either.

grudge/discretization.py
thomasgibson: Good point, I removed it in this commit: https://github.com/inducer/grudge/pull/83/commits/f5f016ef9b81fdbc951ec002130db393d67d0894 (And I don't think you can do either of those things currently, so I'll err on the safe side for now)

grudge/dof_desc.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/a70b9d6f0b904382bed836331f6bb043cb6ea297

grudge/dof_desc.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/a70b9d6f0b904382bed836331f6bb043cb6ea297

grudge/dof_desc.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/a70b9d6f0b904382bed836331f6bb043cb6ea297

grudge/dof_desc.py
thomasgibson: Good call! Done: https://github.com/inducer/grudge/pull/83/commits/a70b9d6f0b904382bed836331f6bb043cb6ea297

grudge/dof_desc.py
```diff
@@ -241,20 +290,43 @@ def fmt(s):
...
+             f"in version 2022.x. Use {_deprecated_names[name]} instead.",
+             DeprecationWarning, stacklevel=2)
+        return globals()[f"{_deprecated_names[name]}"]
+
+    raise AttributeError(f"module {__name__} has no attribute {name}")
+
+
+if sys.version_info < (3, 7):
+    for name in _deprecated_names:
+        name = globals()[f"{_deprecated_names[name]}"]
```
alexfikl, line 329:
```suggestion
        globals()[name] = globals()[_deprecated_names[name]]
```
Untested, but I don't think the previous code actually defined the deprecated variables (?)

thomasgibson: Oops :facepalm: fixing it

grudge/dof_desc.py
thomasgibson: How's this: https://github.com/inducer/grudge/pull/83/commits/4adf94e01bc63ce18d2891723829092b5e40f7b8?

grudge/dof_desc.py
thomasgibson: Nvm not done yet, some issues needing to be sorted out.

grudge/dof_desc.py
thomasgibson: Relevant commit: https://github.com/inducer/grudge/pull/83/commits/c062ec2fe80eb4953e3113abea17a341fe1a25f6

grudge/discretization.py
```diff
@@ -304,33 +313,32 @@ def connection_from_dds(self, from_dd, to_dd):
         else:
             raise ValueError("cannot interpolate from: " + str(from_dd))
 
-    def group_factory_for_quadrature_tag(self, quadrature_tag):
+    def group_factory_for_quadrature_tag(self, discretization_tag):
```
inducer, line 316: Rename/deprecate this method?

grudge/discretization.py
```diff
@@ -304,33 +313,32 @@ def connection_from_dds(self, from_dd, to_dd):
...
 
-        if quadrature_tag is None:
-            quadrature_tag = QTAG_NONE
-
-        return self.quad_tag_to_group_factory[quadrature_tag]
+        return self.quad_tag_to_group_factory[discretization_tag]
 
     @memoize_method
-    def _quad_volume_discr(self, quadrature_tag):
+    def _quad_volume_discr(self, discretization_tag):
```
inducer, line 326: Rename this method?

grudge/dof_desc.py
```diff
@@ -208,29 +265,43 @@ def is_trace(self):
...
+        # `issubclass` will raise an exception whenever
+        # its first argument is not a class.
+        # This can go away once support for strings is dropped
+        # completely.
+        if isinstance(self.discretization_tag, str):
+            return True
+        if issubclass(self.discretization_tag, DISCR_TAG_QUAD):
+            return True
+
+        return False
```
inducer, line 279: I'm not sure I'm loving this fallback to `False` here. I'd prefer a structure where the `else` branch is "I don't know what I'm looking at, help?" (i.e. an error)

grudge/dof_desc.py
```diff
@@ -241,20 +312,43 @@ def fmt(s):
...
+             f"in version 2022.x. Use {_deprecated_names[name]} instead.",
+             DeprecationWarning, stacklevel=2)
+        return globals()[f"{_deprecated_names[name]}"]
+
+    raise AttributeError(f"module {__name__} has no attribute {name}")
+
+
+if sys.version_info < (3, 7):
+    for name in _deprecated_names:
+        globals()[f"{name}"] = globals()[f"{_deprecated_names[name]}"]
```
inducer, line 351:
```suggestion
        globals()[name] = globals()[_deprecated_names[name]]
```

grudge/dof_desc.py
```diff
@@ -241,20 +312,43 @@ def fmt(s):
...
+_deprecated_names = {"QTAG_NONE": "DISCR_TAG_BASE",
+                     "QTAG_MODAL": "DISCR_TAG_MODAL"}
+
+
+def __getattr__(name):
+    if name in _deprecated_names:
+        warn(f"{name} is deprecated and will be dropped "
+             f"in version 2022.x. Use {_deprecated_names[name]} instead.",
+             DeprecationWarning, stacklevel=2)
+        return globals()[f"{_deprecated_names[name]}"]
```
inducer, line 344:
```suggestion
        return globals()[_deprecated_names[name]]
```

grudge/dof_desc.py
```diff
@@ -241,20 +312,43 @@ def fmt(s):
...
+# {{{ Deprecated tags
+
+_deprecated_names = {"QTAG_NONE": "DISCR_TAG_BASE",
+                     "QTAG_MODAL": "DISCR_TAG_MODAL"}
+
+
+def __getattr__(name):
+    if name in _deprecated_names:
+        warn(f"{name} is deprecated and will be dropped "
+             f"in version 2022.x. Use {_deprecated_names[name]} instead.",
```
inducer, line 342:
```suggestion
        warn(f"'{name}' is deprecated and will be dropped "
             f"in version 2022.x. Use '{_deprecated_names[name]}' instead.",
```
(Otherwise it's sometimes hard to tell the boundary between message and names that are under user control.)

grudge/dof_desc.py
```diff
@@ -241,20 +312,43 @@ def fmt(s):
...
         return dd
-    return DOFDesc(dd, quadrature_tag=None)
+    return DOFDesc(dd, discretization_tag=None)
+
+# }}}
+
+
+# {{{ Deprecated tags
+
+_deprecated_names = {"QTAG_NONE": "DISCR_TAG_BASE",
```
inducer, line 335:
```suggestion
_deprecated_name_to_new_name = {"QTAG_NONE": "DISCR_TAG_BASE",
```

inducer (commented):
Thanks! A few (hopefully easy) clean-ups, then this is ready to roll.

inducer:

grudge/dof_desc.py
thomasgibson: Good call! Done in https://github.com/inducer/grudge/pull/83/commits/d778430473a03d93f750e6010a614e9c140d94f7

grudge/dof_desc.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/0dbc12e09731750cc595d5f316f3948b68333207

grudge/dof_desc.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/0dbc12e09731750cc595d5f316f3948b68333207

grudge/dof_desc.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/0dbc12e09731750cc595d5f316f3948b68333207

grudge/dof_desc.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/0dbc12e09731750cc595d5f316f3948b68333207

grudge/discretization.py
thomasgibson: @inducer Sure. Do you want to leave `quad_tag_to_group_factory` argument alone in the class constructor? or should we change that also (to `discr_tag_to_group_factory`)?

inducer: I'd say change it, too. I think we're phasing out the word "quadrature tag", so this makes sense to me.

grudge/discretization.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/83/commits/d4fedd7a1acba4dab4038dac349b36b71a7ca932

grudge/discretization.py
thomasgibson: This function was renamed in this commit: https://github.com/inducer/grudge/pull/83/commits/6c8c3d94fdccd47f6fad605866ff1f37a5cae167, and the `quad_tag_to_group_factory` argument was deprecated in these commits (which includes references to them within the test suite/examples): https://github.com/inducer/grudge/pull/83/commits/fdafb6dbccf084082c076e0f9ad1ef2d12ee71d8, https://github.com/inducer/grudge/pull/83/commits/f239490d1045865f0147cde6b64f2540591043e4

inducer:
:tada: Thanks! I really like how this turned out.

Issue #65, opened 2021-04-11
Grudge keeps a high number of active blocks


nchristensen:
When I run the wave-eager case with a memory pool, I observe the number of active blocks in the memory pool can reach almost 250. Nearly every call to `actx.empty` or `actx.from_numpy` adds another active block. Assuming `pn=3`, for nel_1d=32 (~173,000 elements) and below there is sufficient device memory to execute. For nel_1d=64 (1.5 million elements) the allocations are ~0.25 GB and all of device memory is eventually consumed.

I presume there are a lot of old allocations hanging around that ought to be freed. Perhaps this is related to https://github.com/inducer/pyopencl/issues/450.

inducer:
https://github.com/inducer/pyopencl/issues/450 is unrelated. (Its means of keeping memory alive knows nothing of memory pools.)

nchristensen:
@inducer, I think this assignment is a possible culprit.
https://github.com/inducer/grudge/blob/c03ea0e0523694c251754323b8aa541c36eddd14/grudge/execution.py#L480

`dcoll` is a top level object so any values added to this dictionary hang around until they are removed. After `weak_div()` is called there are twenty 0.25 GB PyOpenCL arrays stored here when running with 1.5 million elements at order 3:

`dict_keys(['discr.dx0_dr0', 'discr.dx1_dr1', 'discr.dx1_dr0', 'discr.dx0_dr1', 'discr.dx2_dr2', 'discr.dx2_dr1', 'discr.dx2_dr0', 'discr.dx1_dr2', 'discr.dx0_dr2', 'discr.pseudoscalar_e0e1e2', 'discr.area_element', 'discr.dr0_dx0', 'discr.dr1_dx0', 'discr.dr2_dx0', 'discr.dr0_dx1', 'discr.dr1_dx1', 'discr.dr2_dx1', 'discr.dr0_dx2', 'discr.dr1_dx2', 'discr.dr2_dx2'])
`

inducer:
Thanks! That makes sense. 18 of them are forward and inverse mapping derivatives (each a 3x3=9 matrix). We only need the "inverse" direction (dr/dx) of these to evaluate the derivatives, so we should only store that. #76 makes that happen, without any run time impact that I was able to measure.

inducer:
cc @thomasgibson (just FYI)

nchristensen:
[...]

PR #85, opened 2021-04-28
Do not cache interior nodes

inducer:
- [ ] Depends on https://github.com/inducer/meshmode/pull/171.

Together with that, it saves storing ~4 big DOF vectors on the GPU.

This goes towards https://github.com/inducer/grudge/issues/65.

cc @nchristensen

### DIFF
--- a/grudge/execution.py
+++ b/grudge/execution.py
@@ -73,7 +73,15 @@ class ExecutionMapper(mappers.Evaluator,
 
     def map_node_coordinate_component(self, expr):
         discr = self.dcoll.discr_from_dd(expr.dd)
-        return thaw(self.array_context, discr.nodes()[expr.axis])
+        return thaw(self.array_context, discr.nodes(
+            # only save volume nodes or boundary nodes
+            # (but not interior ones, which are likely only used once
+            # to compute the normals)
+            cached=(
+                discr.ambient_dim == discr.dim
+                or expr.dd.is_boundary_or_partition_interface()
+                )
+            )[expr.axis])
 
     def map_grudge_variable(self, expr):
         from numbers import Number
--- a/grudge/execution.py
+++ b/grudge/execution.py
@@ -75,8 +75,8 @@ class ExecutionMapper(mappers.Evaluator,
         discr = self.dcoll.discr_from_dd(expr.dd)
         return thaw(self.array_context, discr.nodes(
             # only save volume nodes or boundary nodes
-            # (but not interior ones, which are likely only used once
-            # to compute the normals)
+            # (but not nodes for interior face discretizations, which are likely only
+            # used once to compute the normals)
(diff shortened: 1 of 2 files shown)


grudge/execution.py
```diff
@@ -73,7 +73,15 @@ def map_ones(self, expr):
 
     def map_node_coordinate_component(self, expr):
         discr = self.dcoll.discr_from_dd(expr.dd)
-        return thaw(self.array_context, discr.nodes()[expr.axis])
+        return thaw(self.array_context, discr.nodes(
+            # only save volume nodes or boundary nodes
+            # (but not interior ones, which are likely only used once
+            # to compute the normals)
```
thomasgibson, line 79: Can you elaborate on what you mean by "interior" nodes?

inducer, line 79:
```suggestion
            # (but not nodes for interior face discretizations, which are likely only
            # used once to compute the normals)
```

inducer: How's 85c5c1a?

grudge/execution.py
```diff
@@ -73,7 +73,15 @@ def map_ones(self, expr):
 
     def map_node_coordinate_component(self, expr):
         discr = self.dcoll.discr_from_dd(expr.dd)
-        return thaw(self.array_context, discr.nodes()[expr.axis])
+        return thaw(self.array_context, discr.nodes(
+            # only save volume nodes or boundary nodes
+            # (but not nodes for interior face discretizations, which are likely only
+            # used once to compute the normals)
```
thomasgibson, line 79: Ah okay!

Though I wondering if this will be an issue for entropy stable DG. For example, when doing the volume flux differencing, we need to compute values on the surface of elements too. Which means we have to interpolate to the face nodes. But I guess if those interpolations are already done, there's no need to hang on to the surface nodes though? So maybe it's okay

thomasgibson (commented):
LGTM, I just have a clarifying question (thinking ahead about entropy stable DG). I think though that this won't be an issue, just wanted to know your thoughts

grudge/execution.py
inducer: This all deals with a global node array (hence the memory consumption concern). I don't believe those are used in constructing the interpolation (just uses reference nodes).

inducer:
@thomasgibson I'm trying to understand whether this is good to merge or not from your POV. Could you respond/resolve the conversation above and/or indicate how you feel about merging?

grudge/execution.py
thomasgibson: OK, thanks!

thomasgibson (approved):

grudge/execution.py
inducer: I'm guessing this is good to resolve, too?

PR #87, opened 2021-04-29
Vectorize local_grad and weak_local_grad

majosm:
Open question: should these return an array of arrays or a 2D array?

### DIFF
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -148,7 +148,7 @@ def local_grad(dcoll, vec, *, stack=True):
     """
     if isinstance(vec, np.ndarray):
         grad = obj_array_vectorize(
-                lambda el: local_grad(dcoll, el), vec)
+                lambda el: local_grad(dcoll, el, stack=stack), vec)
         if stack:
             return np.stack(grad, axis=0)
         else:
@@ -242,7 +242,7 @@ def weak_local_grad(dcoll, *args, stack=True):
 
     if isinstance(vec, np.ndarray):
         grad = obj_array_vectorize(
-                lambda el: weak_local_grad(dcoll, dd, el), vec)
+                lambda el: weak_local_grad(dcoll, dd, el, stack=stack), vec)
         if stack:
             return np.stack(grad, axis=0)
         else:
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -142,6 +142,10 @@ def local_grad(dcoll, vec):
     :arg vec: a :class:`~meshmode.dof_array.DOFArray`
     :returns: an object array of :class:`~meshmode.dof_array.DOFArray`\ s
     """
+    if isinstance(vec, np.ndarray):
+        return obj_array_vectorize(
+                lambda el: local_grad(dcoll, el), vec)
+
     return _bound_grad(dcoll)(u=vec)
 
 
@@ -223,6 +227,10 @@ def weak_local_grad(dcoll, *args):
         dd, vec = args
     else:
         raise TypeError("invalid number of arguments")
+
+    if isinstance(vec, np.ndarray):
+        return obj_array_vectorize(
+                lambda el: weak_local_grad(dcoll, dd, el), vec)
 
     return _bound_weak_grad(dcoll, dd)(u=vec)
 
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -135,24 +135,24 @@ def _bound_grad(dcoll):
     return bind(dcoll, sym.nabla(dcoll.dim) * sym.Variable("u"), local_only=True)
 
 
-def local_grad(dcoll, vec, *, stack=True):
+def local_grad(dcoll, vec, *, nested=False):
     r"""Return the element-local gradient of the volume function represented by
     *vec*.
 
     :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
         `~meshmode.dof_array.DOFArray`
-    :arg stack: return a multidimensional object array instead of nested arrays
-        if *vec* is non-scalar
+    :arg nested: return nested object arrays instead of a single multidimensional
+        array if *vec* is non-scalar
     :returns: an object array (possibly nested) of
         :class:`~meshmode.dof_array.DOFArray`\ s
     """
     if isinstance(vec, np.ndarray):
         grad = obj_array_vectorize(
-                lambda el: local_grad(dcoll, el, stack=stack), vec)
-        if stack:
-            return np.stack(grad, axis=0)
-        else:
+                lambda el: local_grad(dcoll, el, nested=nested), vec)
+        if nested:
             return grad
+        else:
+            return np.stack(grad, axis=0)
 
     return _bound_grad(dcoll)(u=vec)
 
@@ -217,7 +217,7 @@ def _bound_weak_grad(dcoll, dd):
             local_only=True)
 
 
-def weak_local_grad(dcoll, *args, stack=True):
+def weak_local_grad(dcoll, *args, nested=False):
     r"""Return the element-local weak gradient of the volume function
     represented by *vec*.
 
@@ -227,8 +227,8 @@ def weak_local_grad(dcoll, *args, stack=True):
         Defaults to the base volume discretization if not provided.
     :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
         `~meshmode.dof_array.DOFArray`
-    :arg stack: return a multidimensional object array instead of nested arrays
-        if *vec* is non-scalar
+    :arg nested: return nested object arrays instead of a single multidimensional
+        array if *vec* is non-scalar
     :returns: an object array (possibly nested) of
         :class:`~meshmode.dof_array.DOFArray`\ s
     """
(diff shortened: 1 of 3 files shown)


inducer:
> Open question: should these return an array of arrays or a 2D array?

I think both are justifiable, and I'm grumpy that we have to choose. I'm OK with either:
* Unconditionally return an array of array for now until actual use teaches us that something different is needed.
* Make it an option. `flatten=True` or  so (?) for 2D array.

grudge/op.py
```diff
@@ -142,6 +142,10 @@ def local_grad(dcoll, vec):
     :arg vec: a :class:`~meshmode.dof_array.DOFArray`
     :returns: an object array of :class:`~meshmode.dof_array.DOFArray`\ s
     """
+    if isinstance(vec, np.ndarray):
```
inducer, line 145: Add a test?

MTCam:
> > Open question: should these return an array of arrays or a 2D array?
> 
> I think both are justifiable, and I'm grumpy that we have to choose. I'm OK with either:
> 
> * Unconditionally return an array of array for now until actual use teaches us that something different is needed.
> * Make it an option. `flatten=True` or  so (?) for 2D array.

2D array return is required for *MIRGE-Com* where we `np.stack` the returned array to get the 2D representation. So it'd be great if it did return just what we need - although I am not sure how that could potentially break things in other developments.

Actually, now that I think about it.. since this is a new capability returning a 2D array shouldn't break anything outstanding.  If anyone wants to get back at a flat rep, they could always convert it or just use `obj_array_vectorize` like is currently done in *MIRGE*.

inducer:
> If anyone wants to get back at a flat rep, they could always convert it or just use obj_array_vectorize like is currently done in MIRGE.

To fix terminology: What I mean when I said "flat" was the 2D array, whereas, to me, the "array of arrays" thing you might get from `obj_array_vectorize` is "nested".

MTCam:
Another thought ...

I think it would be reasonable to expect that if `q` is the system state vector and one called `r = weak_local_grad(q)` or its ilk, that they'd get an `r` that has an `ndim-vector` for each component of `q`, and so a shape of  `r.shape=(num_equations, ndim)` can be immediately understood by the user.  

However, since it took me a little while to piece together what was wrong with the structure that i was getting back from something like `obj_arr_vec(grad(q))`, I can say from experience that (for me) dealing with the nested representation was unexpected, and harder.

One additional merit of the 2D rep is that it works with `cross_rank_trace_pairs` now? :)

grudge/op.py
```diff
@@ -224,6 +228,10 @@ def weak_local_grad(dcoll, *args):
     else:
         raise TypeError("invalid number of arguments")
 
+    if isinstance(vec, np.ndarray):
```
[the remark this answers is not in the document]
majosm: I added a test (for both vectorized and non-vectorized; I couldn't find anything that was already testing `op.local_grad`/`op.weak_local_grad`). Does this look ok?

Also, any ideas why:

1) my strong form implementation doesn't seem to work?
2) flake8 is yelling at me? (I tried indenting those lines differently, but so far I haven't been able to make the warning go away.)

majosm:
I added a parameter to control whether `local_grad`/`weak_local_grad` returns a 2D array or an array of arrays. I called the parameter `stack` for now, because `flatten` seemed potentially confusing to me (I usually think of flattening as producing a 1D array). I'm open to changing it though, especially if there's a precedent for using `flatten` to refer to "nested -> 2D".

inducer:
> `flatten`

How about a `nested` flag? I think I like that better terminology-wise than either `flatten` or `stack`.

inducer:
As for flake8, I think it's not liking the comment before the close-paren. This seems to make it happy (and it seems to make at least a residual bit of sense):
```diff
diff --git a/test/test_op.py b/test/test_op.py
index 965e1ce..c6945ff 100644
--- a/test/test_op.py
+++ b/test/test_op.py
@@ -123,7 +123,7 @@ def test_gradient(actx_factory, dim, form, order, vectorize, stack):
                     -  # noqa: W504
                     get_flux(op.interior_trace_pair(dcoll, u))
                     # No boundary flux terms because u_ext == u_int == 0
-                    )
+                    + 0)
                 )
         elif form == "weak":
             grad_u = op.inverse_mass(dcoll,
@@ -133,7 +133,7 @@ def test_gradient(actx_factory, dim, form, order, vectorize, stack):
                     dd_allfaces,
                     get_flux(op.interior_trace_pair(dcoll, u))
                     # No boundary flux terms because u_ext == u_int == 0
-                    )
+                    + 0)
                 )
 
         if vectorize:
```
This does have a runtime cost though. Another option would be to move the comments above the interior fluxes.

inducer:
The pylint issue seems specific to that particular version of pylint, https://github.com/inducer/ci-support/pull/12.

test/test_op.py
```diff
@@ -0,0 +1,168 @@
...
+                -  # noqa: W504
+                op.face_mass(dcoll,
+                    dd_allfaces,
+                    # Note: no boundary flux terms here because u_ext == u_int == 0
+                    get_flux(sym.TracePair(dd_intfaces,
+                        interior=u_intfaces,
+                        exterior=u_intfaces))
+                    -  # noqa: W504
+                    get_flux(op.interior_trace_pair(dcoll, u)))
+                )
```
thomasgibson, line 126:
```suggestion
                - op.local_grad(dcoll, u, nested=nested)
                +  # noqa: W504
                op.face_mass(dcoll,
                    dd_allfaces,
                    # Note: no boundary flux terms here because u_ext == u_int == 0
                    get_flux(sym.TracePair(dd_intfaces,
                        interior=u_intfaces,
                        exterior=u_intfaces))
                    -  # noqa: W504
                    get_flux(op.interior_trace_pair(dcoll, u)))
                )
```
I think there is a sign error here?

thomasgibson: Huh. I noticed you flipped the signs consistently (even in the weak case). Nevermind.

majosm:
I tweaked `local_div`/`weak_local_div` to work with both nested and non-nested arrays and added some tests for them.

My strong form tests for both `div` and `grad` currently aren't working. I disabled them for the moment, but I welcome anyone to take a look and tell me what I'm doing wrong. 🙂 

I'm still getting the pylint error with the current pylint version (2.8.2). Looks like a false positive. This example script reproduces it:

```python
import numpy as np
from pytools.obj_array import make_obj_array

a = make_obj_array([np.array([1,2,3]), np.array([4,5,6])])
b = -np.stack(a, axis=0)
```

Inspecting `b` shows:

```python
type(b)=<class 'numpy.ndarray'>
b.shape=(2, 3)
b.dtype=dtype('int64')
b=array([[-1, -2, -3],
       [-4, -5, -6]])
```

but pylint says:

```
script.py:5:4: E1130: bad operand type for unary -: list (invalid-unary-operand-type)
```

I guess I'll just annotate the places where this warning pops up?

majosm:
@inducer Yeah, removing the `inverse_mass` fixed the strong form tests. I think this is ready for review now.

test/test_op.py
```diff
@@ -0,0 +1,310 @@
...
+                op.inverse_mass(dcoll,
+                    op.face_mass(dcoll,
+                        dd_allfaces,
+                        # Note: no boundary flux terms here because
+                        # u_ext == u_int == 0
+                        get_flux(sym.TracePair(dd_intfaces,
+                            interior=u_intfaces,
+                            exterior=u_intfaces))
+                        -  # noqa: W504
+                        get_flux(op.interior_trace_pair(dcoll, u))))
```
inducer, line 256: As above for the strong-form flux.

test/test_op.py
```diff
@@ -0,0 +1,310 @@
...
```
inducer, line 127: I'm not sure this test is meaningful, since the strong-form flux ultimately comes down to the jump between elements, and since you're starting with a continuous function, I'm almost certain the flux is zero here.

inducer (commented):
Thanks! A few questions surrounding the strong-form fluxes in the tests below.

test/test_op.py
majosm: This is true. I removed the flux terms and it still works.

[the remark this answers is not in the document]
majosm: I think I have both terms there, right? (Whether they're implemented correctly is another question...)

Moot point now though, I guess -- the fluxes have been removed.

test/test_op.py
```diff
@@ -0,0 +1,286 @@
...
+            normal = thaw(actx, op.normal(dcoll, dd))
+            flux = u_tpair.avg @ normal
+            return op.project(dcoll, dd, dd_allfaces, flux)
+
+        dd_allfaces = DOFDesc("all_faces")
+
+        if form == "strong":
+            div_u = (
+                op.local_div(dcoll, u)
+                # No flux terms because u is smooth
```
inducer, line 232:
```suggestion
                # No flux terms because u doesn't have inter-el jumps
```

test/test_op.py
```diff
@@ -0,0 +1,286 @@
...
+            else:
+                flux = u_avg * normal
+            return op.project(dcoll, dd, dd_allfaces, flux)
+
+        dd_allfaces = DOFDesc("all_faces")
+
+        if form == "strong":
+            grad_u = (
+                op.local_grad(dcoll, u, nested=nested)
+                # No flux terms because u is smooth
```
inducer, line 115:
```suggestion
                # No flux terms because u doesn't have inter-el jumps
```

inducer (approved):
Thanks! LGTM. I'll apply these two phrasing changes.

PR #91, opened 2021-05-10
Fix wave example blow ups and check norms

thomasgibson:
Similar to https://github.com/inducer/meshmode/pull/173/


inducer:
Thanks!

PR #98, opened 2021-05-24
Add normals/nodes to dcoll

thomasgibson:
cc @MTCam @inducer

For discussion based on this thread: https://github.com/illinois-ceesd/mirgecom/pull/360#discussion_r638097941

### DIFF
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -615,7 +615,7 @@ class DiscretizationCollection:
 
         :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
             Defaults to the base volume discretization.
-        :returns: an object array of :class:`~meshmode.dof_array.DOFArray`\ s
+        :returns: an object array of frozen :class:`~meshmode.dof_array.DOFArray`\ s
         """
         if dd is None:
             dd = DD_VOLUME
@@ -626,7 +626,7 @@ class DiscretizationCollection:
         r"""Get the unit normal to the specified surface discretization, *dd*.
 
         :arg dd: a :class:`~grudge.dof_desc.DOFDesc` as the surface discretization.
-        :returns: an object array of :class:`~meshmode.dof_array.DOFArray`\ s.
+        :returns: an object array of frozen :class:`~meshmode.dof_array.DOFArray`\ s.
         """
         from arraycontext import freeze
         from grudge.geometry import normal
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -32,8 +32,12 @@ THE SOFTWARE.
 from pytools import memoize_method
 
 from grudge.dof_desc import (
-    DISCR_TAG_BASE, DISCR_TAG_MODAL,
-    DTAG_BOUNDARY, DOFDesc, as_dofdesc
+    DD_VOLUME,
+    DISCR_TAG_BASE,
+    DISCR_TAG_MODAL,
+    DTAG_BOUNDARY,
+    DOFDesc,
+    as_dofdesc
 )
 
 import numpy as np  # noqa: F401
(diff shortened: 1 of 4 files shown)


grudge/discretization.py
```diff
@@ -519,6 +521,38 @@ def order(self):
...
+        :returns: an object array of :class:`~meshmode.dof_array.DOFArray`\ s
+        """
+        if dd is None:
+            return self._volume_discr.nodes()
+        else:
+            return self.discr_from_dd(dd).nodes()
+
+
+    @memoize_method
+    def normal(self, dd):
```
MTCam, line 540: Will the memoization here do anything to prevent update if the mesh moves?  Just curious.

thomasgibson:
I am leaning towards delaying https://github.com/illinois-ceesd/mirgecom/pull/360 until https://github.com/inducer/grudge/pull/74 in merged in Grudge. The reason being is that this will be (hopefully soon) the right place to start.

grudge/discretization.py
thomasgibson: That's a good question. I don't think so. This would have to be inferred I guess if the `nodes` are changing.

inducer: See https://github.com/inducer/meshmode/issues/123 for some related discussion. IMO, generally the safest approach would be to discard and rebuild the whole collection when it changes, while working to make that a (reasonably) cheap operation.

grudge/discretization.py
```diff
@@ -519,6 +521,38 @@ def order(self):
         from pytools import single_valued
         return single_valued(egrp.order for egrp in self._volume_discr.groups)
 
+    # {{{ Discretization-specific geometric properties
+
+    def nodes(self, dd=None):
```
inducer, line 526: Add tests?

inducer:
I agree that this should wait until after #74.

inducer:

grudge/discretization.py
MTCam: Will this discarding and rebuilding be ok when the structures surface regresses at every substep?  Even in the absence of regression (caused by erosion, or burning); if we have real coupling there will be deformation displacements - which definitely occur at the substep level; that is, every time the surface load (i.e. the fluid pressure) changes.  This approach seems like a lot of tear-down-setup overhead.

inducer: The thing is that making new objects is (relatively) cheap. If it's a new mesh, then by far the most expensive thing will be remapping to a new mesh. If it's smooth deformations, then IMO the cost will be dominated by updating/recomputing area elements and mapping derivatives. Making a new `DiscretizationCollection`, by comparison, is dirt cheap.

inducer: [That constructor](https://github.com/inducer/grudge/blob/0a7b2fb2e4e170809b55e0c3c193ae1dbc4a61a4/grudge/discretization.py#L74-L162) currently does a bit more work than it needs to, but I think that can be addressed. #104

grudge/discretization.py
```diff
@@ -601,6 +608,33 @@ def order(self):
...
+        r"""Return the nodes of a discretization specified by *dd*.
+
+        :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
+            Defaults to the base volume discretization.
+        :returns: an object array of :class:`~meshmode.dof_array.DOFArray`\ s
```
inducer, line 618:
```suggestion
        :returns: an object array of frozen :class:`~meshmode.dof_array.DOFArray`\ s
```

grudge/discretization.py
```diff
@@ -601,6 +608,33 @@ def order(self):
...
+        if dd is None:
+            dd = DD_VOLUME
+        return self.discr_from_dd(dd).nodes()
+
+    @memoize_method
+    def normal(self, dd):
+        r"""Get the unit normal to the specified surface discretization, *dd*.
+
+        :arg dd: a :class:`~grudge.dof_desc.DOFDesc` as the surface discretization.
+        :returns: an object array of :class:`~meshmode.dof_array.DOFArray`\ s.
```
inducer, line 629:
```suggestion
        :returns: an object array of frozen :class:`~meshmode.dof_array.DOFArray`\ s.
```

inducer (approved):
Thanks for working on this! LGTM, I'll apply these.

PR #101, opened 2021-05-24
Docs: Stop linking to source code from docs

inducer:
As of https://github.com/inducer/grudge/pull/74, there might actually be something resembling adequate documentation so that we don't need the sad excuse of "just go look at the source".

@thomasgibson What do you think?

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -15,7 +15,6 @@ extensions = [
     "sphinx.ext.doctest",
     "sphinx.ext.intersphinx",
     "sphinx.ext.mathjax",
-    "sphinx.ext.viewcode",
     "sphinx_copybutton",
 ]
 


thomasgibson:
Any reason not to do both?

MTCam:
> Any reason not to do both?

fwiw, i *really* like both, but am not opposed to getting trained off of that.

thomasgibson:
> > Any reason not to do both?
> 
> fwiw, i _really_ like both, but am not opposed to getting trained off of that.

I like both too. I don't really see any reason to drop it. Why not link to the source code for curious readers?

inducer:
:shrug: OK. Nvm then. Thanks for chiming in with your thoughts.

PR #105, opened 2021-05-28
Fix nodal min

thomasgibson:


thomasgibson:
@inducer I think something in `dagrt` broke the example in `old_symbolic`.

inducer:
Thanks for finding the issue with the `min`. I'll work on getting the dagrt fusion thing fixed up.

inducer:
Dagrt problem is being fixed in #106, which will hopefully merge soon.

thomasgibson:
Thanks @inducer !

grudge/op.py
```diff
@@ -1122,7 +1122,7 @@ def nodal_min(dcoll: DiscretizationCollection, dd, vec):
     # NOTE: Must retain a way to do local reductions
     actx = vec.array_context
     return reduce(lambda acc, grp_ary: actx.np.minimum(acc, actx.np.min(grp_ary)),
-                  vec, -np.inf)
```
inducer, line 1125: Wanna add a test? :)

inducer (approved):
(I approve, test or no.)

grudge/op.py
inducer: (Setting to merge, tests could happen in a separate PR.)

PR #108, opened 2021-05-30
Update time-step utility functions

thomasgibson:
This PR adds time-stepping utility functions

Also updates all the examples


thomasgibson:
@inducer Unfortunately, none of the examples work with the value the timestep estimate is producing. I'm trying to understand why

inducer:
> @inducer Unfortunately, none of the examples work with the value the timestep estimate is producing. I'm trying to understand why

What's more important (to my mind) is that the scaling is right. So:
* Add a fudge factor until stable.
* Then change *h* (up and down). Things should stay stable.
* Then bump up the fudge factor slightly to assess the sharpness of the estimate. (Things should go unstable.)
* Then repeat the exercise with *p*.
* Then repeat the exercise with changing aspect ratio.
* Lastly, cast this whole procedure as a unit test.
* A few steps of advection should suffice to assess stability.

thomasgibson:
> What's more important (to my mind) is that the scaling is right. So:
> 
> * Add a fudge factor until stable.
> * Then change _h_ (up and down). Things should stay stable.
> * Then bump up the fudge factor slightly to assess the sharpness of the estimate. (Things should go unstable.)
> * Then repeat the exercise with _p_.
> * Then repeat the exercise with changing aspect ratio.
> * Lastly, cast this whole procedure as a unit test.
> * A few steps of advection should suffice to assess stability.

Unfortunately, the fudge factor is problem-dependent, meaning that I can get advection examples working (the exception being the surface example...) but the Maxwell/wave problems still blow up. I may have a bug somewhere, or perhaps this isn't quite as robust as we thought.

inducer:
I just spent some time playing around with this, and I think it's probably as good as it as has any right to be. First of all, advection is an odd bird, and not that great of a test: it propagates information in only one direction. Wave and Maxwell propagate in all directions. Also, for Maxwell, the version currently in the repo uses less upwinding than wave:
https://github.com/inducer/grudge/blob/596482cbc13be794a7cef80058260cf4930b37bd/examples/maxwell/cavities.py#L77
...because. To make it a fair comparison, I cranked this up to 1 (i.e. "full upwinding")

Playing around in 3D, the fudge factor I got for the wave equation was .23, while Maxwell got .27 (for shorter runs... so maybe not quite as sensitive a test).

IMO, what matters is that this captures the dependency on mesh and reference discretization, so that you can change those without having to re-fiddle the fudge factor. From a cursory test, it seems to do that fairly well.

thomasgibson:
@inducer I'm happy with this now, and I've managed to get every single example working with minimal tweaking. The only thing to note is that, for the `examples/wave-op-X.py` examples, I need to decrease the scaling coefficient in the non-geometric factor (0.6667 -> 0.45) :shrug:. The other wave examples were happy with the default 2/3 factor. Not super sure why, but I've made the function accept an optional scaling parameter so that people can tweak it fairly easily.

grudge/models/__init__.py
```diff
@@ -36,16 +38,24 @@ class Operator:
 class HyperbolicOperator(Operator):
     """A base class for hyperbolic Discontinuous Galerkin operators."""
 
-    def max_eigenvalue(self, t, fields, discr):
-        raise NotImplementedError
+    @abstractmethod
+    def max_eigenvalue(self, t, fields, dcoll):
+        """Return an upperbound on the maximal eigenvalue for the operator."""
```
inducer, line 43: While the characteristic speed are typically found as an eigenvalues, I think a rename to `max_characteristic_vel` would be an improvement.

inducer (commented):
A few things ahead of a full review, to give you a chance to address them.

examples/maxwell/cavities.py
```diff
@@ -86,18 +86,17 @@ def cavity_mode(x, t=0):
...
         return maxwell_operator.operator(t, w)
 
-    if mesh.dim == 2:
-        dt = 0.004
-    elif mesh.dim == 3:
-        dt = 0.002
+    # if mesh.dim == 2:
+    #     dt = 0.004
+    # elif mesh.dim == 3:
+    #     dt = 0.002
```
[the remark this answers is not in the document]
thomasgibson: Done in https://github.com/inducer/grudge/pull/108/commits/5723aa378c9cf660880f2cd5d6bf58229d96e021!

examples/wave/var-propagation-speed.py
```diff
@@ -100,10 +100,12 @@ def source_f(actx, dcoll, t=0):
...
 
-    if mesh.dim == 2:
-        dt = 0.04 * 0.3
-    elif mesh.dim == 3:
-        dt = 0.02 * 0.1
+    # if mesh.dim == 2:
+    #     dt = 0.04 * 0.3
+    # elif mesh.dim == 3:
+    #     dt = 0.02 * 0.1
+
```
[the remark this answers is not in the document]
thomasgibson: Done in https://github.com/inducer/grudge/pull/108/commits/5723aa378c9cf660880f2cd5d6bf58229d96e021!

examples/wave/wave-min.py
```diff
@@ -55,10 +55,10 @@ def main(write_output=False, order=4):
...
             nelements_per_axis=(16,)*dims)
 
-    if mesh.dim == 2:
-        dt = 0.04
-    elif mesh.dim == 3:
-        dt = 0.02
+    # if mesh.dim == 2:
+    #     dt = 0.04
+    # elif mesh.dim == 3:
+    #     dt = 0.02
```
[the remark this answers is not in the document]
thomasgibson: Done in https://github.com/inducer/grudge/pull/108/commits/5723aa378c9cf660880f2cd5d6bf58229d96e021!

examples/wave/wave-min-mpi.py
```diff
@@ -118,7 +118,7 @@ def source_f(actx, dcoll, t=0):
     )
 
     # FIXME
-    # dt = wave_op.estimate_rk4_timestep(dcoll, fields=fields)
+    dt = wave_op.estimate_rk4_timestep(dcoll, fields=fields)
```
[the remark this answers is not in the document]
thomasgibson: Done! https://github.com/inducer/grudge/pull/108/commits/5b35e3ecd1315fc5ea2fc8ca0e9ced99b95ad11e

grudge/models/__init__.py
thomasgibson: Good idea, I like that suggestion: https://github.com/inducer/grudge/pull/108/commits/6998f984145151df07d85e1e8eaf981cc59092b6

examples/wave/wave-op-var-velocity.py
```diff
@@ -126,6 +129,16 @@ def rk4_step(y, t, h, f):
     return y + h/6*(k1 + 2*k2 + 2*k3 + k4)
 
 
+def estimate_rk4_timestep(dcoll, c, scaling=None):
```
inducer, line 132: Not a huge fan of this scaling parameter. Users might as well multiply the output by something.

grudge/models/__init__.py
```diff
@@ -36,16 +38,26 @@ class Operator:
...
+        dt_factor = \
+            (dt_non_geometric_factor(dcoll, scaling=dt_scaling)
+             * dt_geometric_factor(dcoll))
+
+        mpi_comm = dcoll.mpi_communicator
+        if mpi_comm is None:
+            return dt_factor * (1 / max_lambda)
+
+        return (1 / mpi_comm.allreduce(max_lambda, op=MPI.MAX)) \
+            * mpi_comm.allreduce(dt_factor, op=MPI.MIN)
```
inducer, line 63: I think we'll also want a local-only version of this, i.e. something without the reduction baked in. `estimate_rk4_timestep_loc`?

x-ref: #113

inducer, line 63: Shouldn't one reduction suffice for this?

grudge/models/advection.py
```diff
@@ -84,8 +84,8 @@ def __init__(self, dcoll, v, inflow_u=None, flux_type="central"):
     def weak_flux(self, u_tpair):
         return advection_weak_flux(self.dcoll, self.flux_type, u_tpair, self.v)
 
-    def max_eigenvalue(self, t=None, fields=None, discr=None):
-        return np.linalg.norm(self.v)
+    def max_characteristic_velocity(self, t=None, fields=None, dcoll=None):
+        return op.norm(self.dcoll, self.v, 2)
```
inducer, line 88: This needs to be a rank-local norm, otherwise you're double-reducing.

grudge/models/em.py
```diff
@@ -332,7 +332,7 @@ def get_eh_subset(self):
         """
         return 6*(True,)
 
-    def max_eigenvalue_expr(self):
+    def max_characteristic_vel_expr(self):
```
inducer, line 335: Why have this (instead of just `max_characteristic_velocity`)? I think this may have made sense in the symbolic world, but not so much any more.

test/test_grudge.py
```diff
@@ -850,7 +850,7 @@ def analytic_sol(x, t=0):
         def rhs(t, w):
             return maxwell_operator.operator(t, w)
 
-        dt = 0.002
```
inducer, line 853: Could you add some tests that verify the scaling properties of the `dt_*_factor` routines? (probably best in a separate file)

* Increasing the order superlinearly decreases the non-geometric factor.
* Uniform scaling of the mesh reduces the geometric factor by exactly the scale factor.
* Single-axis scaling of the mesh reduces the geometric factor by a factor related to the scale factor

grudge/dt_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    .. math::
+
+        C\operatorname{min}_i\left( \Delta r_i \right),
+
+    where :math:`\Delta r_i` denotes the distance between two distinct
+    nodes on the reference element, and :math:`C > 0` is a scaling
+    constant.
+
+    :arg scaling: a :class:`float` denoting the scaling factor. By default,
+        the constant is set to 2/3.
```
inducer, line 60: I'd kind of feel better if we removed this argument and let users scale the returned time step. Otherwise we end up documenting the 2/3, which feels kinda  weird. (But I'm open to discussion here.)

grudge/dt_utils.py
```diff
@@ -0,0 +1,206 @@
...
+        # the reference element and is equidistant from each vertex
+        if grp.order == 0:
+            assert nnodes == 1
+            min_delta_rs.append(
+                np.linalg.norm(
+                    nodes[0] - grp.mesh_el_group.vertex_unit_coordinates()[0]
+                )
+            )
+        else:
+            min_delta_rs.append(
```
inducer, line 88: Provide a reference for this approach.

grudge/dt_utils.py
```diff
@@ -0,0 +1,206 @@
...
+    cell_vols = DOFArray(
+        actx,
+        data=tuple(
+            actx.call_loopy(cv_prg(),
+                            cell_vol=cv_i)["result"]
+
+            for vgrp, cv_i in zip(volm_discr.groups,
+                                  actx.np.fabs(cell_vols))
+        )
+    )
```
inducer, line 160: Could do this as an einsum with a post-multiply.

grudge/dt_utils.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Specifically, the inradius for each element is computed using the following
+    formula for simplicial cells (triangles/tetrahedra):
+
+    .. math::
+
+        r_{in} = \frac{d V}{\sum_{i=1}^{N_{faces}} F_i},
+
+    where :math:`d` is the topological dimension, :math:`V` is the cell volume,
+    and :math:`F_i` are the areas of each face of the cell.
```
inducer, line 112: Should this be documented? As in, are we committing to this approach?

inducer, line 112: Provide a reference for the formula.

grudge/dt_utils.py
```diff
@@ -0,0 +1,206 @@
...
+                                vgrp.mesh_el_group.nfaces,
+                                vgrp.nelements,
+                                afgrp.nunit_dofs
+                            ))["result"]
+
+            for vgrp, afgrp, face_ae_i in zip(volm_discr.groups,
+                                              face_discr.groups,
+                                              actx.np.fabs(face_areas))
+        )
+    )
```
inducer, line 204: Could be done as an einsum with a post-mutiply.

inducer (commented):
Thanks! A few thoughts below.

inducer:

grudge/models/advection.py
thomasgibson: Technically all of our norms are rank local right now. Not sure what you want me to do here.

inducer: See #113 for context. Basically, anytime we use `op.norm` is a  risk until we know whether that's supposed to be local or not.

> Technically all of our norms are rank local right now.

The "right now" is exactly where I'm having a problem. Suppose that changes, then this needs to change, too. There's no need for that churn. We should just fix the interface so that user code can tell whether they'll be using a local or distributed norm. Then we don't need to change all the existing usage sites again. That's why, IMO, #113 is fairly urgent.

grudge/dt_utils.py
thomasgibson: > Should this be documented?

100% always yes.

> are we committing to this approach?

Well, I would say it's worth committing to this (it's certainly better than having _nothing_ ). There are objections to using `h_min_from_volume`, so this seemed like a reasonable alternative. What are the other alternatives here? Inradius seems to be a well-known and accepted mesh size parameter for polygonal meshes.

There are more elaborate constructions I'm sure we could entertain, but I'd rather not and just use something simple that works reasonably well.

grudge/dt_utils.py
thomasgibson: This is based on Hesthaven and Warburton, which is heuristically determined.

@MTCam this means in your code, you'll need to add a factor of `2/3` in front of `non_geometric_factor * geometric_factor` in your timestep calculation. Does that feel okay with you?

grudge/dt_utils.py
thomasgibson: https://github.com/inducer/grudge/pull/108/commits/297377e60ffee1a65499b25b3819eda29c3f1f43

grudge/dt_utils.py
thomasgibson: https://github.com/inducer/grudge/pull/108/commits/297377e60ffee1a65499b25b3819eda29c3f1f43

grudge/dt_utils.py
inducer: Fair enough. We're not really at liberty to change what this routine does without risking breakage, so we might as well document what it does on the inside.

grudge/dt_utils.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/108/commits/021fd898e939eebde9dc65e9700e9f9363a59f14

grudge/dt_utils.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/108/commits/021fd898e939eebde9dc65e9700e9f9363a59f14

grudge/models/em.py
thomasgibson: Done! https://github.com/inducer/grudge/pull/108/commits/92607dfbc85b2e1274ec42c781976c989bf1ec0b

examples/wave/wave-op-var-velocity.py
thomasgibson: Got rid of these in https://github.com/inducer/grudge/pull/108/commits/b2a8a35c8366c156d30b9a409373b9b505398d16

grudge/models/__init__.py
thomasgibson: Marked to address this in a separate PR: https://github.com/inducer/grudge/pull/108/commits/f18a36970bd18dfe7e890d19813d97b9f0aed5ab

grudge/models/advection.py
thomasgibson: Marked to address this in a separate PR: https://github.com/inducer/grudge/pull/108/commits/f18a36970bd18dfe7e890d19813d97b9f0aed5ab

grudge/dt_utils.py
thomasgibson: Marked to address this in a separate PR: https://github.com/inducer/grudge/pull/108/commits/f18a36970bd18dfe7e890d19813d97b9f0aed5ab

test/test_grudge.py
thomasgibson: @inducer I've added a refinement test for the geometric scaling factors, but I'm not quite sure how to craft a test for the non-geometric factors. Got any hints?

inducer: I was thinking vary the order between like 1 and 8 and check `lower_bound <= nongeom * n**2 <= upper_bound`. Kinda crude, but better than nothing.

thomasgibson: Thanks! I think I got a decent test set up here: https://github.com/inducer/grudge/pull/108/commits/267fa878f7b3b93d5b5fe392480a5cd202cd9f9d

thomasgibson:
@inducer I'm at a loss with the CI failures :confused: The difference between the last (successful) commit and the current (failing) one is a single unit test. The error messages in https://github.com/inducer/grudge/pull/108/checks?check_run_id=2747071808#step:3:858 indicate something going on with the MPI test but I'm unable to reproduce it locally

inducer:
* I restarted the tests, let's see if that buys us anything (probably not).
* Next, it might be related to hwloc, so we'll need to check versions.
* If that doesn't help, we can ssh into the CI via https://github.com/marketplace/actions/debugging-with-tmate.

grudge/op.py
```diff
@@ -1198,6 +1202,25 @@ def elementwise_sum(dcoll: DiscretizationCollection, *args):
...
+
+def elementwise_integral(dcoll: DiscretizationCollection, dd, vec):
+    """Numerically integrates a function represented by a
+    :class:`~meshmode.dof_array.DOFArray` of degrees of freedom in
+    each element of a discretization, given by *dd*.
+
+    :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray`
+    :returns: a :class:`~meshmode.dof_array.DOFArray` containing the
+        elementwise integral if *vec*.
```
inducer, line 1214:
```suggestion
        elementwise integral of *vec*, represented as a constant function on each
        cell.
```

test/test_dt_utils.py
```diff
@@ -0,0 +1,129 @@
...
+            op.nodal_min(dcoll, "vol", dt_geometric_factors(dcoll))
+        )
+
+    # Crude estimate, factors should behave like 1/N**2
+    factors = np.asarray(factors)
+    geom_factors = np.asarray(geom_factors)
+    lower_bounds = 1/(np.asarray(degrees)**2)
+    upper_bounds = geom_factors*lower_bounds
+
+    assert lower_bounds.all() <= factors.all() <= upper_bounds.all()
```
inducer, line 118: I don't think this works. `.all()` returns an array of Booleans. Likewise, chained inequalities also only work on Python scalars, not arrays. I tried a quick fix, but the inequalities are violated.

inducer (commented):
This looks great. Only two minor things, and then is ready to go in.

test/test_dt_utils.py
thomasgibson: @inducer Ah whoops. I've updated it: https://github.com/inducer/grudge/pull/108/commits/77c2a0a256a69866b0b7233b91000c70942757b3

inducer:
Thanks! LGTM, in it goes.

PR #110, opened 2021-06-02
Fix, test complex norms

inducer:
Noticed this last night when playing with #108. The Maxwell cavities example was returning complex-valued norms, which seemed odd... :)

cc @thomasgibson


thomasgibson:
Thanks for catching that! I meant to address this at some point but was preoccupied with timestepping :laughing:. Thanks for fixing this

PR #115, opened 2021-06-03
Simplify WADG mass inverse

### LINKED ISSUE #114: WADG mass-matrix inverse: correctness, excessive flop count
Missed this when reviewing #74.

https://github.com/inducer/grudge/blob/30db06b0192f961b989c3bbf45c0f310f098cd9e/grudge/op.py#L767-L785

If `Ndof` is the number of DOFs, then the code above uses `Ndof**4` flops per element, when `Ndof**2` would suffice. What should happen here?

* First of all, I'm not sure the implementation is correct. This computes `inv(ref_M) @ ref_M` as the leading term, which is an identity. To be fair, this came from the original implementation:
  https://github.com/inducer/grudge/blob/30db06b0192f961b989c3bbf45c0f310f098cd9e/grudge/symbolic/mappers/__init__.py#L644-L645
* Then, any matrices that *can* be premultiplied, should be premultiplied.
* Last, any einsums involved should at most have one reduction variable. (Our current einsum isn't smart enough to do anything else efficiently.)

cc @thomasgibson

inducer:
Closes #114

cc @alexfikl

### DIFF
--- a/grudge/symbolic/mappers/__init__.py
+++ b/grudge/symbolic/mappers/__init__.py
@@ -591,9 +591,6 @@ class GlobalToReferenceMapper(CSECachingMapperMixin, IdentityMapper):
         self.ambient_dim = dcoll.ambient_dim
         self.dim = dcoll.dim
 
-        volume_discr = dcoll.discr_from_dd(dof_desc.DD_VOLUME)
-        self.use_wadg = not all(grp.is_affine for grp in volume_discr.groups)
-
     map_common_subexpression_uncached = \
             IdentityMapper.map_common_subexpression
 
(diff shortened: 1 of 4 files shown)


inducer:
Big question: How confident are we in our tests?

alexfikl:
Well,
* The only [actual test](https://github.com/inducer/grudge/blob/30db06b0192f961b989c3bbf45c0f310f098cd9e/test/test_grudge.py#L268-L325) just checks that it is indeed a reasonable inverse.
* The [surface advection example](https://github.com/inducer/grudge/blob/30db06b0192f961b989c3bbf45c0f310f098cd9e/examples/advection/surface.py) also uses the mass inverse in [here](https://github.com/inducer/grudge/blob/30db06b0192f961b989c3bbf45c0f310f098cd9e/grudge/models/advection.py#L358). At least from what I remember, that seems to be stable without going crazy with the CFL, but @thomasgibson probably has a better feel for it with the recent dt work.

It could probably use a couple more tests!

thomasgibson:
I would hope that https://github.com/inducer/grudge/blob/30db06b0192f961b989c3bbf45c0f310f098cd9e/test/test_grudge.py#L268-L325 would be sufficient, but perhaps another good test would be to mimic that test, but using the warped mesh (like in `examples/geometry.py`). That guy is non-affine

thomasgibson:
@inducer I hope you're okay with me pushing a short unit test on warped geometry!

thomasgibson (approved):

inducer:
> @inducer I hope you're okay with me pushing a short unit test on warped geometry!

Sure, that's totally fine. Could that geometry just be a case in the other test? Or am I missing an aspect there?

grudge/symbolic/mappers/__init__.py
```diff
@@ -641,12 +641,9 @@ def imd(rst):
...
-                    op.RefMassOperator(dd_in, dd_out)(
+                return (
                         1.0/jac_in * op.RefInverseMassOperator(dd_in, dd_out)(
-                            self.rec(expr.field))
-                            )
-                    )
+                            self.rec(expr.field)))
             else:
                 return op.RefInverseMassOperator(dd_in, dd_out)(
                         1/jac_in * self.rec(expr.field))
```
alexfikl, line 649: Unife the two branches here too? Like in `grudge.op`

inducer: 217bfc5

inducer:
@thomasgibson In 74abc6b, I folded your new test into the existing one. Hope that's OK.

thomasgibson:
@inducer Totally fine with me!

PR #120, opened 2021-06-06
Retire redundant examples

thomasgibson:
Retires the wave-op and wave-min examples, since the mpi counterparts cover the exact same test case


inducer:
Thx!

Issue #119, opened 2021-06-05
Timestep estimate might be unnecessarily conservative


inducer:
This code:

https://github.com/inducer/grudge/blob/57d90c902b50fa757f7d6436d9834dbec1cfaf72/grudge/models/__init__.py#L56-L59

takes the min of the geometric factors and the min of the non-geometric factors over all groups and then multiplies them. Technically, I think it would suffice to multiply them together (still obtaining a nodal quantity) and then taking the min. Reducing down to per-group quantities and then multiplying and then taking the min would also work. This is a bit hypothetical for now, since most use is single-group, but it might bite us in the future.

I missed this while reviewing #108.

cc @thomasgibson @MTCam

thomasgibson:
I think that's a very fair point. So the non geometric factor will be a single value per element group. Is there a safe way to broadcast the `dt_non_geometric_factor`s, an array of shape `(ngroups,)` and with a `DOFArray` that doesn't mess with the object types?

inducer:
Sure! Something like this ought to work:
```
DOFArray(actx, tuple(
    nongeo_fac * grp_ary
    for nongeo_fac, grp_ary in zip(dt_non_geometric_factors, dof_ary)))
```
(Remember that each sub-array in the `DOFArray` corresponds exactly to a group.)

PR #121, opened 2021-06-06
Return non-geometric factors for all groups

thomasgibson:


grudge/models/__init__.py
```diff
@@ -49,13 +49,22 @@ def max_characteristic_velocity(self, t, fields, dcoll):
...
+            actx,
+            data=(
+                cng * geo_facts
+                for cng, geo_facts in zip(dt_non_geometric_factors(dcoll),
+                                          dt_geometric_factors(dcoll))
+            )
+        )
 
-        return dt_factor * (1 / max_lambda)
+        return op.nodal_min(dcoll, "vol", dt_factors * (1 / max_lambda))
```
thomasgibson, line 70: @MTCam Sorry for the switch-up on this. But could I get your opinion on this? For reference: https://github.com/inducer/grudge/issues/119

inducer, line 70: For fluids, I feel like the multiplication by `1/max_lambda` should probably also be inside the `nodal_min`.

inducer, line 70: I'd honestly feel better if this were a function rather than a method. Otherwise, it starts forcing objects into places where they kind of have no business being. If that means killing the `HyperbolicOperator` class, that's probably for the better. (mirgecom uses functions for its operators, and I think that's by and large a good thing.)

MTCam: Is there only one geometric factor per group?  I don't think I understood that before.  In order to get a truly cell-local estimate for dt/cfl, we'll need cell-local geom/nongeom factors.  

I want to explain a little bit what my concern is... In our meshes, we have very tight spacing in the boundary layer, and likely we will refine in recirculation zones where combustion will be initiated.  The flow velocities in these very fine mesh areas should, in general, be much slower.

A common approach is to estimate the global max dt using max wavespeed (e.g. MAX_mesh(|v| + |c|))  and the MIN_mesh(h).  This makes a lot of sense usually (simple geom, simple flow) - but for our cases with complex flow and geometric features - we should be a little smarter, or we end up with an unnecessarily conservative dt (potentially costing us *months* in compute time).  Our PlasCom2 simulations run for weeks-on-end to produce prediction results.  MirgeCom will require more. These seemingly slight differences accumulate a lot over our relatively long runtimes, and have significant impact on the required compute time to do our prediction.

The first stab at doing something better is kind of naive and obvious, but effective. We just do the *cell-local* values for *h*, and the *cell-local* wavespeeds, to get a *cell-local* dt.  The max/min consideration is deferred until the end where we choose the DT by taking a min over cells, or the max CFL by maxing over cells.  MIRGE-Com needs the back-end to support this approach.

All of that said, ANY estimate of the sim cfl or the max dt is going to work for us.  It is not something that ever needs to be precise. The user will take any numbers coming out of this as a *guide* only; the actual stability analysis for any given simulation is an empirical operation.

thomasgibson: It is :)

thomasgibson: > Is there only one geometric factor per group? I don't think I understood that before. In order to get a truly cell-local estimate for dt/cfl, we'll need cell-local geom/nongeom factors.

Nope, there is one _non geometric_ factor per group. This factor depends on the type of reference cell (triangle, tetrahedron, prim, etc) and the polynomial degree. The geometric factor is an actual measure of the cell diameter, and so will be different for each cell in the mesh (unless you have an evenly spaced, regular mesh).

What this PR does is that it will return all non geometric factors (by group) to you rather than take the minimum over all groups. This line here:
```python
dt_factors = DOFArray(
            actx,
            data=tuple(
                cng * geo_facts
                for cng, geo_facts in zip(dt_non_geometric_factors(dcoll),
                                          dt_geometric_factors(dcoll))
            )
        )
```
produces a dof array with the CFL grid factors at each nodal point (since `dt_geometric_factors` is a dof array of size `(nelem, ndofs)`. The difference here is that the scaling by the non-geometric factor will be different across all groups now (if you have different mesh elements/poly order).

MTCam: Got it @thomasgibson. Thanks for that explanation!

thomasgibson: I haven't completely removed this function, but is this closer to what you're thinking: https://github.com/inducer/grudge/pull/121/commits/a3659999b0e5dcfa57de21c1090e4a706b675629?

inducer: Yep, that looks excellent!

thomasgibson: I think this: https://github.com/inducer/grudge/pull/121/commits/a3659999b0e5dcfa57de21c1090e4a706b675629 cleans things up a bit and adds a more convenient way to get this information

grudge/dt_utils.py
```diff
@@ -45,22 +57,58 @@
...
+    where :math:`\operatorname{min}\left(\Delta r_i\right)` is the minimum
+    node distance on the reference cell (see :func:`dt_non_geometric_factors`),
+    :math:`r_D` is the inradius of the physical cell
+    (see :func:`dt_geometric_factors`), and :math:`|c|` is the local
+    magnitude of the characteristic wavespeed.
+
+    :arg wavespeed: a number or :class:`~meshmode.dof_array.DOFArray` for the
+        characteristic wavespeed.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` containing the local
+        timstep estimation at each nodal location.
```
inducer, line 78:
```suggestion
    :returns: a :class:`~meshmode.dof_array.DOFArray` containing an estimate
        of the maximal stable time step for explicit time integration estimation
        at each nodal location.

    .. note::

        While a prediction of stability is only meaningful in relation to a given
        time integrator with a known stability region, the estimate provided here
        is not intended to be specific to any one time integrator, though the
        stability region of standard four-stage, fourth-order Runge-Kutta
        methods has been used as a guide. Any concrete time integrator will
        likely require scaling of the value returned by this routine.
```

grudge/dt_utils.py
```diff
@@ -45,22 +57,58 @@
...
+        \Delta t_{loc} = \operatorname{min}\left(\Delta r_i\right)\frac{r_D}{|c|}
+
+    where :math:`\operatorname{min}\left(\Delta r_i\right)` is the minimum
+    node distance on the reference cell (see :func:`dt_non_geometric_factors`),
+    :math:`r_D` is the inradius of the physical cell
+    (see :func:`dt_geometric_factors`), and :math:`|c|` is the local
+    magnitude of the characteristic wavespeed.
+
+    :arg wavespeed: a number or :class:`~meshmode.dof_array.DOFArray` for the
+        characteristic wavespeed.
```
inducer, line 76: I wonder if we should make the wavespeed optional, to allow use of the returned quantity as a replacement of `h/N^2`, e.g. in the formula given by @MTCam in this comment: https://github.com/illinois-ceesd/mirgecom/pull/368#discussion_r641814288

inducer (commented):
Thanks! This looks great. Two minor things below.

grudge/dt_utils.py
MTCam: That would be great.  In general, and imo much to the credit of code reviews in our current effort, I've been trying to avoid making functions take arguments that are used for scaling only:

e.g., I find:
`a = fac * f(x, y, z)` 

... is superior to:
`a = f(x, y, z, fac)`
(with the scaling on the inside)

The former is easier to read and understand immediately what is happening.  The current function is no exception.  The physics user developers will understand as much as they need from:

`dt = grudge.dt_facs/wavespeed`, while something like: `grudge.estimate_local_timestep(discr, wavespeed)` is a bit more opaque, requiring a bit more digging to figure out how the timestep is estimated.   

This clarity advantage is in addition to the thing that @inducer mentions above.  I would like to write something like:

`dt = grudge.dt_facs/(wavespeed + mu/grudge.dt_facs)` 

... for a quick and dirty viscous timestep estimate.

thomasgibson: So I've added @inducer's suggestion above https://github.com/inducer/grudge/pull/121/commits/cd2740380c16884db521ee86fc1e98c9608eabc8. But do you both (@MTCam + @inducer) think this should be renamed to `dt_factors` or something like that and always return just the dt scaling factor? I'm trying to get a sense on what you both prefer. Currently, to do what @MTCam wants, you'd basically write:
```python
dt_factors = estimate_local_timestep(dcoll)

dt = dt_factors/(wavespeed + mu/dt_factors)
```

MTCam: Thanks a bunch, @thomasgibson! I defer to your judgements about the naming of it.  One thing you *could* do is provide a public function something like: `get_characteristic_length_scale` and then your `estimate_local_timestep` can use that, or users can directly use the `get_characteristic_length` or whatever if they want.

inducer: I like `characteristic_length`. I'll leave it up to you whether to keep `estimate_local_timestep` as a wrapper, but my gut feeling is that
```
dt = characteristic_length(...) / wavespeed
```
is about the same length and perhaps clearer than `estimate_local_timestep(..., wavespeed)`.

MTCam: That's shiny!  Thanks @thomasgibson!

grudge/dt_utils.py
```diff
@@ -45,22 +57,59 @@
...
+
+        While a prediction of stability is only meaningful in relation to a given
+        time integrator with a known stability region, the lengthscale provided here
+        is not intended to be specific to any one time integrator, though the
+        stability region of standard four-stage, fourth-order Runge-Kutta
+        methods has been used as a guide. Any concrete time integrator will
+        likely require scaling of the values returned by this routine.
+    """
+    return DOFArray(
+        dcoll._setup_actx,
```
inducer, line 85: The setup array context should never escape into user expressions, unless you're keen on very difficult to debug synchronization bugs.

Some general rules:
* if the user did not explicitly supply an array context, a routine should only ever return frozen data.
* Thawed data (i.e. data that's attached to an array context) should never be memoized.

Clearly, this concept needs a better explanation. I'll work on that.

x-ref: https://github.com/inducer/arraycontext/issues/19

grudge/dt_utils.py
```diff
@@ -45,22 +57,59 @@
...
+        data=tuple(
+            # Scale each group array of geometric factors by the
+            # corresponding group non-geometric factor
+            cng * geo_facts
+            for cng, geo_facts in zip(dt_non_geometric_factors(dcoll),
+                                      dt_geometric_factors(dcoll))
```
inducer, line 91: `dt_geometric_factors` also returns unfrozen data in the same way. I propose that we fix this in this PR.

x-ref: https://github.com/inducer/arraycontext/issues/19

inducer (commented):
Thanks! Some doc issues and a pretty significant freeze/thaw bug.

inducer:
@thomasgibson I changed `HyperbolicOperator.max_characteristic_velocity` to permit local data to be returned.

inducer:
Hmm, looks like that drove the surface advection case unstable...

grudge/dt_utils.py
inducer: https://github.com/inducer/arraycontext/pull/22

MTCam:
> Hmm, looks like that drove the surface advection case unstable...

Curious, and mildly alarming. Do you happen to already know what the bottom line is as far as the estimate for DT, and which cells are the limiting cells in terms of wavespeeds and characteristic length?  

It *could* be that the DT estimate is now closer to the actual stability limit, and the problem should be run at a CFL somewhat less than 1.

inducer:
> Curious, and mildly alarming.

Not immensely alarming, at least from my perspective. Nobody claimed that these particular stability bounds would work for surfaces... so I was kind of surprised when they did. In addition, I don't know that the formula @thomasgibson found for the incircle radius is at all valid on surfaces.

inducer:
Added a fudge factor in 813621c and a warning in 1092ac2.

thomasgibson:
@inducer Thanks for the chat earlier! I think this is ready now for a final look.

inducer:
> @inducer Can you explain why? Was it wrong before?

It wasn't wrong, it just wasn't as sharp as it could be, for much the same reason as #119. Taking the global max of characteristic velocities and the min of the discretization bits yields a more conservative bound than just taking the minimum once.

grudge/dt_utils.py
```diff
@@ -45,22 +57,65 @@
 from pytools import memoize_on_first_arg
 
 
+def characteristic_lengthscales(dcoll: DiscretizationCollection) -> DOFArray:
```
inducer, line 60: I feel like this would be a candidate for memoization, moreso than `dt_geometric_factors`.

grudge/models/wave.py
```diff
@@ -198,27 +198,28 @@ class VariableCoefficientWeakWaveOperator(HyperbolicOperator):
...
-    def __init__(self, dcoll, c, source_f=None,
+    def __init__(self, actx, dcoll, c, source_f=None,
             flux_type="upwind",
             dirichlet_tag=BTAG_ALL,
             dirichlet_bc_f=0,
             neumann_tag=BTAG_NONE,
             radiation_tag=BTAG_NONE):
+        """
+        :arg c: a thawed (with *actx*) :class:`~meshmode.dof_array.DOFArray`
+            representing the propogation speed of the wave.
```
inducer, line 211: Should follow the advice from https://github.com/inducer/arraycontext/pull/22/: Just use the array context from `c`.

inducer (commented):
Thanks! Two more small things, and this is good to go from my end.

inducer:
#124 might address those CI failures, currently set to automerge.

grudge/models/__init__.py
```diff
@@ -54,7 +54,7 @@ def estimate_rk4_timestep(self, actx, dcoll, t=None, fields=None):
         from grudge.dt_utils import characteristic_lengthscales
         import grudge.op as op
 
-        wavespeeds = self.max_characteristic_velocity(t, fields, dcoll)
+        wavespeeds = self.max_characteristic_velocity(actx, t, fields)
```
inducer, line 57: Either:

* Update the interface spec of `max_characteristic_velocity`, or
* Make passing `fields` mandatory, so that you can extract the array context from that, or
* get the array context from one or the other, make both optional, with one required.

grudge/dt_utils.py
```diff
@@ -46,21 +58,65 @@
 
 
 @memoize_on_first_arg
-def dt_non_geometric_factor(dcoll: DiscretizationCollection, dd=None) -> float:
-    r"""Computes the non-geometric scale factor following [Hesthaven_2008]_,
-    section 6.4:
+def characteristic_lengthscales(dcoll: DiscretizationCollection) -> DOFArray:
```
inducer, line 61: At the dev meeting this morning, @MTCam pointed out that, usually, the very next thing folks will do with the result from this is some type of arithmetic, so it's awkward if they have to thaw it. As a result, this should probably take an array context and return a thawed result.

As far as the memoization is concerned, you can simply take this function, rename it to say `_characteristic_lengthscales_inner`, and provide a "thawing wrapper".

inducer (commented):
Two further comments. Sorry about the piecemeal way in which they're showing u.

grudge/models/__init__.py
thomasgibson: I've updated the interface a bit here: https://github.com/inducer/grudge/pull/121/commits/697cc7bc37871415e7b62a3e0a7d8a8066ff256c I figured since all these operators will eventually be rewritten in a more functional form, it's probably not worth investing too much into the operator interfaces. So I just opted for a quick-and-easy solution

grudge/dt_utils.py
thomasgibson: Done here: https://github.com/inducer/grudge/pull/121/commits/ec7ca8345bfacd50be6d24c54e1bfc15f29fcff6

inducer:
:tada: LGTM. Thanks!

Issue #103, opened 2021-05-26
Is DD_VOLUME misnamed?


inducer:
Given that the "volume" discretization might describe a surface.

Along with the `"vol"` argument to `as_dofdesc`. 

cc @alexfikl

alexfikl:
Is there some language in geometric algebra or exterior calculus that we can borrow for this? Mostly just a thought, since nothing comes to mind..

PR #122, opened 2021-06-07
Refactor DiscretizationCollection constructor

thomasgibson:
This PR removes all the work that the current constructor for `DiscretizationCollection` is doing for setup. The proposed interface for creating these discretization collections is provided by the function `make_discretization_collection`.

### DIFF
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -112,7 +112,8 @@ class DiscretizationCollection:
             discretization.
         :arg dist_boundary_connections: A dictionary whose keys denote the
             partition group index and map to the appropriate face connections
-            for distributed boundaries, if any.
+            for distributed boundaries, if any. See
+            :func:`set_up_distributed_communication` for more information.
         :arg mpi_communicator: An (optional) MPI communicator.
         """
         self._setup_actx = array_context
@@ -736,6 +737,8 @@ def set_up_distributed_communication(
         comm=None) -> Dict[int, DiscretizationConnection]:
     """Constructs a mapping from parallel boundary partition and the relevant
     discretization connections to that boundary. Used for distributed runs.
+    Connected partitions are determined by
+    :func:`meshmode.distributed.get_connected_partitions`.
 
     :arg base_discr: A :class:`meshmode.discretization.Discretization`
         object for the base (:class:`grudge.dof_desc.DISCR_TAG_BASE`)
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -35,6 +35,7 @@ from pytools import memoize_method
 
 from grudge.dof_desc import (
     DD_VOLUME,
+    DD_VOLUME_MODAL,
     DISCR_TAG_BASE,
     DISCR_TAG_MODAL,
     DTAG_BOUNDARY,
@@ -82,7 +83,7 @@ class DiscretizationCollection:
 
     def __init__(self, array_context: ArrayContext, mesh: Mesh,
                  discr_tag_to_group_factory,
-                 volume_discr,
+                 discr_from_dd,
                  dist_boundary_connections,
                  mpi_communicator=None):
         """
@@ -95,17 +96,20 @@ class DiscretizationCollection:
             to be carried out, or *None* to indicate that operations with this
             discretization tag should be carried out with the standard volume
             discretization.
-        :arg volume_discr: A :class:`meshmode.discretization.Discretization`
-            object for the base (:class:`grudge.dof_desc.DISCR_TAG_BASE`)
-            volume discretization.
+        :arg discr_from_dd: A mapping from discretization tags
+            from a :class:`grudge.dof_desc.DOFDesc`
+            to a :class:`meshmode.discretization.Discretization`.
+            At minimum, this should include a base discretization given by
+            the dof descriptor :class:`grudge.dof_desc.DD_VOLUME`.
         :arg dist_boundary_connections: A dictionary whose keys denote the
             partition group index and map to the appropriate face connections
             for distributed boundaries, if any.
         :arg mpi_communicator: An (optional) MPI communicator.
         """
         self._setup_actx = array_context
+        self._mesh = mesh
         self.discr_tag_to_group_factory = discr_tag_to_group_factory
-        self._volume_discr = volume_discr
+        self._discr_from_dd = discr_from_dd
 
         # NOTE: Can be removed when symbolics are completely removed
         # {{{ management of discretization-scoped common subexpressions
@@ -184,15 +188,16 @@ class DiscretizationCollection:
         """
         dd = as_dofdesc(dd)
 
+        if dd in self._discr_from_dd:
+            return self._discr_from_dd[dd]
+
         discr_tag = dd.discretization_tag
 
         if discr_tag is DISCR_TAG_MODAL:
             return self._modal_discr(dd.domain_tag)
 
         if dd.is_volume():
-            if discr_tag is not DISCR_TAG_BASE:
-                return self._discr_tag_volume_discr(discr_tag)
-            return self._volume_discr
+            return self._discr_tag_volume_discr(discr_tag)
 
         if discr_tag is not DISCR_TAG_BASE:
             no_quad_discr = self.discr_from_dd(DOFDesc(dd.domain_tag))
@@ -322,7 +327,7 @@ class DiscretizationCollection:
                 from meshmode.discretization.connection import \
                         make_same_mesh_connection
                 to_discr = self._discr_tag_volume_discr(to_discr_tag)
-                from_discr = self._volume_discr
+                from_discr = self._discr_from_dd[DD_VOLUME]
                 return make_same_mesh_connection(self._setup_actx, to_discr,
                             from_discr)
 
@@ -355,7 +360,7 @@ class DiscretizationCollection:
         from meshmode.discretization import Discretization
 
         return Discretization(
-            self._setup_actx, self._volume_discr.mesh,
+            self._setup_actx, self._mesh,
             self.group_factory_for_discretization_tag(discretization_tag)
         )
 
@@ -409,7 +414,7 @@ class DiscretizationCollection:
     def _boundary_connection(self, boundary_tag):
         return make_face_restriction(
             self._setup_actx,
-            self._volume_discr,
+            self._discr_from_dd[DD_VOLUME],
             self.group_factory_for_discretization_tag(DISCR_TAG_BASE),
             boundary_tag=boundary_tag
         )
@@ -422,7 +427,7 @@ class DiscretizationCollection:
     def _interior_faces_connection(self):
         return make_face_restriction(
             self._setup_actx,
-            self._volume_discr,
+            self._discr_from_dd[DD_VOLUME],
             self.group_factory_for_discretization_tag(DISCR_TAG_BASE),
             FACE_RESTR_INTERIOR,
 
@@ -453,7 +458,7 @@ class DiscretizationCollection:
     def _all_faces_volume_connection(self):
         return make_face_restriction(
             self._setup_actx,
-            self._volume_discr,
+            self._discr_from_dd[DD_VOLUME],
             self.group_factory_for_discretization_tag(DISCR_TAG_BASE),
             FACE_RESTR_ALL,
 
@@ -468,29 +473,29 @@ class DiscretizationCollection:
     @property
     def dim(self):
         """Return the topological dimension."""
-        return self._volume_discr.dim
+        return self._discr_from_dd[DD_VOLUME].dim
 
     @property
     def ambient_dim(self):
         """Return the dimension of the ambient space."""
-        return self._volume_discr.ambient_dim
+        return self._discr_from_dd[DD_VOLUME].ambient_dim
 
     @property
     def real_dtype(self):
         """Return the data type used for real-valued arithmetic."""
-        return self._volume_discr.real_dtype
+        return self._discr_from_dd[DD_VOLUME].real_dtype
 
     @property
     def complex_dtype(self):
         """Return the data type used for complex-valued arithmetic."""
-        return self._volume_discr.complex_dtype
+        return self._discr_from_dd[DD_VOLUME].complex_dtype
 
     @property
     def mesh(self):
         """Return the :class:`meshmode.mesh.Mesh` over which the discretization
         collection is built.
         """
-        return self._volume_discr.mesh
+        return self._mesh
 
     def empty(self, array_context: ArrayContext, dtype=None):
         """Return an empty :class:`~meshmode.dof_array.DOFArray` defined at
@@ -501,7 +506,7 @@ class DiscretizationCollection:
             vector of dtype :attr:`complex_dtype`. If
             *None* (the default), a real vector will be returned.
         """
-        return self._volume_discr.empty(array_context, dtype)
+        return self._discr_from_dd[DD_VOLUME].empty(array_context, dtype)
 
     def zeros(self, array_context: ArrayContext, dtype=None):
         """Return a zero-initialized :class:`~meshmode.dof_array.DOFArray`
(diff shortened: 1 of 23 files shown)


grudge/discretization.py
```diff
@@ -636,6 +626,133 @@ def normal(self, dd):
     # }}}
 
 
+def make_discretization_collection(
+        array_context: ArrayContext, mesh: Mesh,
```
inducer, line 630: In the long run, that second argument should permit:
* A mesh
* A discretization
* A dictionary from identifiers (like wall/fluid) to either of the above. cc @majosm 

Maybe make document it as positional-only so that we're not tied to its name.

grudge/discretization.py
```diff
@@ -80,7 +80,10 @@ class DiscretizationCollection:
 
     def __init__(self, array_context: ArrayContext, mesh: Mesh,
                  order=None,
-                 discr_tag_to_group_factory=None, mpi_communicator=None,
+                 discr_tag_to_group_factory=None,
+                 volume_discr=None,
```
inducer, line 84: "*The* volume discretization" will be less of a thing once we support multi-physics. So maybe prepare for that by using a dictionary from identifiers to discretizations.

inducer, line 84: I remember this came up recently: For surface discretizations, "volume" is  a bit of a misnomer.

inducer (commented):
Thanks for starting to think about this. Some thoughts from a quick scroll.

grudge/discretization.py
thomasgibson: Good point, I've pushed some changes that remove all references to `volume_discr`

thomasgibson: related: https://github.com/inducer/grudge/issues/103 -- we'll probably need to introduce a `BASE_DISCR` tag or something like that

grudge/discretization.py
thomasgibson: @inducer Could you provide a sketch of what you mean?

inducer: There are two points I'm trying to make here:
* The first one is identical to the comment below. Multi-physics means there's not *the* volume discretization. Instead, there's a map from identifiers to discretizations, with some way to also describe connectivity.
* IMO there's a legitimate reason to want to create a discretization collection if you already *have* a `Discretization`. I.e. it seems like overreach to force the user to provide a mesh and then converting it to a discretization for them.

thomasgibson: > * IMO there's a legitimate reason to want to create a discretization collection if you already _have_ a `Discretization`. I.e. it seems like overreach to force the user to provide a mesh and then converting it to a discretization for them.

@inducer I cannot think of a current example (using grudge) where any example/experiment explicitly constructs a `Discretization`; it's always exclusively through `DiscretizationCollection`. I agree we should permit flexible arguments to the function. But the heart of my question is more or less what do you think is a good placeholder for the moment? I don't plan to solve this question in this PR.

inducer: Sticking with a `Mesh` is alright, we can easily disambiguate by type. Just be forward-looking with naming.

thomasgibson: @inducer Would appreciate your thoughts on https://github.com/inducer/grudge/pull/122/commits/74eb9bb9472ef67e279d7fc02bed71d29711cb8b when you have time. I have to admit, I wasn't entirely sure how to handle this since we have no concrete use cases/examples. So if you had something else in mind, just let me know and I'll update.

inducer:
Is this ready for review from your end? (I ask because of the ambiguous state: no review requested, work still ongoing, but also not marked as draft)

grudge/discretization.py
```diff
@@ -93,64 +98,20 @@ def __init__(self, array_context: ArrayContext, mesh: Mesh,
...
             discretization tag should be carried out with the standard volume
             discretization.
+        :arg discr_from_dd: A mapping from discretization tags
+            from a :class:`grudge.dof_desc.DOFDesc`
+            to a :class:`meshmode.discretization.Discretization`.
+            At minimum, this should include a base discretization given by
+            the dof descriptor :class:`grudge.dof_desc.DD_VOLUME`.
+        :arg dist_boundary_connections: A dictionary whose keys denote the
+            partition group index and map to the appropriate face connections
+            for distributed boundaries, if any.
```
inducer, line 108: How do the keys denote these things? Document their data type.

grudge/discretization.py
```diff
@@ -93,64 +98,20 @@ def __init__(self, array_context: ArrayContext, mesh: Mesh,
             to be carried out, or *None* to indicate that operations with this
             discretization tag should be carried out with the standard volume
             discretization.
+        :arg discr_from_dd: A mapping from discretization tags
+            from a :class:`grudge.dof_desc.DOFDesc`
+            to a :class:`meshmode.discretization.Discretization`.
+            At minimum, this should include a base discretization given by
+            the dof descriptor :class:`grudge.dof_desc.DD_VOLUME`.
```
inducer, line 105: I think the documentation of what this is needs to be expanded a bit, to avoid some misunderstandings that I had when first reading this, mainly on how this relates to the `discr_from_dd` method. (Maybe a different name would also help... `vol_discr_from_dd`?)

- The `discr_from_dd` method can be used to retrieve all sorts of stuff (boundary discretizations and whatnot). This `discr_from_dd` dictionary cannot be used to supply "pre-made" versions of those. (It might be attractive to allow that? not sure) Said another way, this should set an expectation for what can and cannot be part of this mapping.
- This is treated as read-only, i.e. additional discretizations made by `discr_from_dd` aren't added to this.

grudge/discretization.py
```diff
@@ -79,10 +84,10 @@ class DiscretizationCollection:
     """
 
     def __init__(self, array_context: ArrayContext, mesh: Mesh,
```
inducer, line 86: It might be worthwhile removing the constructor from the publicly documented interface.

grudge/discretization.py
```diff
@@ -79,10 +84,10 @@ class DiscretizationCollection:
...
-                 order=None,
-                 discr_tag_to_group_factory=None, mpi_communicator=None,
-                 # FIXME: `quad_tag_to_group_factory` is deprecated
-                 quad_tag_to_group_factory=None):
+                 discr_tag_to_group_factory,
+                 discr_from_dd,
+                 dist_boundary_connections,
+                 mpi_communicator=None):
         """
```
inducer, line 91: This version is definitely no longer compatible with code that calls the `DiscretizationCollection` constructor directly. That's OK for grudge because you've fixed all the call sites, and it's OK for mirgecom because it only uses `EagerDiscretization` to which you've added the backward compatibility code. But I'm not sure it's OK for all other users? (cc @alexfikl)

Alternatively, the compat code could live here.

grudge/discretization.py
```diff
@@ -636,17 +563,166 @@ def normal(self, dd):
...
+        array_context: ArrayContext,
+        discr_context,
+        order=None,
+        discr_tag_to_group_factory=None,
+        mpi_communicator=None) -> DiscretizationCollection:
+    """Construct a discretization collection given an array context *array_context*
+    and user-provided discretization parameters.
+
+    :arg discr_context: A :class:`~meshmode.mesh.Mesh` object to define
+        the discretization collection over.
```
inducer, line 581: Not sure I love the name. In pytential, @alexfikl used "places": https://documen.tician.de/pytential/symbolic.html in a similar role.

grudge/discretization.py
```diff
@@ -636,17 +563,166 @@ def normal(self, dd):
...
+    :arg discr_tag_to_group_factory: A mapping from discretization tags
+        (typically one of: :class:`grudge.dof_desc.DISCR_TAG_BASE`,
+        :class:`grudge.dof_desc.DISCR_TAG_MODAL`, or
```
inducer, line 584: Is overriding the meaning of "modal" allowed?

grudge/discretization.py
```diff
@@ -636,17 +563,166 @@ def normal(self, dd):
...
+                )
+
+            discr_tag_to_group_factory[DISCR_TAG_BASE] = \
+                default_simplex_group_factory(base_dim=mesh.dim, order=order)
+
+    # Modal discr should always comes from the base discretization
+    discr_tag_to_group_factory[DISCR_TAG_MODAL] = \
+        _generate_modal_group_factory(
+            discr_tag_to_group_factory[DISCR_TAG_BASE]
+        )
```
inducer, line 629: This should error if `DISCR_TAG_MODAL` was supplied, instead of silently overwriting. The docs specifically allow supplying modal.

grudge/discretization.py
```diff
@@ -636,17 +563,166 @@ def normal(self, dd):
...
+
+    volume_discr = Discretization(
+        array_context, mesh,
+        discr_tag_to_group_factory[DISCR_TAG_BASE]
+    )
+
+    modal_vol_discr = Discretization(
+        array_context, mesh,
+        discr_tag_to_group_factory[DISCR_TAG_MODAL]
+    )
```
inducer, line 642: This is created "eagerly" here. What's the cost of that? Is it necessary?

grudge/discretization.py
```diff
@@ -636,17 +563,166 @@ def normal(self, dd):
...
+        discr_from_dd=discr_from_dd,
+        dist_boundary_connections=dist_boundary_connections,
+        mpi_communicator=mpi_communicator
+    )
+
+
+def set_up_distributed_communication(
+        array_context: ArrayContext, mesh: Mesh,
+        volume_discr,
+        discr_tag_to_group_factory, comm=None) -> dict:
```
inducer, line 667: `
```suggestion
        discr_tag_to_group_factory, comm=None) -> Dict[key_type, value_type]:
```

inducer (commented):
Thanks! A few comments below.

inducer:

grudge/discretization.py
alexfikl: I have some grudge code sitting around, but it needs a lot of changes to move away from the deprecated symbolic stuff, so I'm fine with making a more breaking change here.

grudge/discretization.py
alexfikl: In particular, the [GeometryCollection](https://documen.tician.de/pytential/symbolic.html#pytential.GeometryCollection) class in there. (The name was used in the code before introducing the GeometryCollection, so I just named it like that to stay consistent)

grudge/discretization.py
thomasgibson: Snuck this in https://github.com/inducer/grudge/pull/122/commits/6bdc25ca42df6731fc2f632403f1a64d64c75f28

grudge/discretization.py
thomasgibson: Good catch! Removed in https://github.com/inducer/grudge/pull/122/commits/fa9a8b904134bd46c7ed6ceef4b656d98c76d467

grudge/discretization.py
thomasgibson: I've updated this in https://github.com/inducer/grudge/pull/122/commits/fa9a8b904134bd46c7ed6ceef4b656d98c76d467; I've opted to only provide the group factor when it's not provided by the user

grudge/discretization.py
thomasgibson: I think this is addressed in https://github.com/inducer/grudge/commit/fa9a8b904134bd46c7ed6ceef4b656d98c76d467. A default modal factory is provided whenever it is not explicitly provided by the user (in a similar way to how there is a default simplex group factor)

grudge/discretization.py
thomasgibson: I decided to do away with the redundant discretizations altogether and streamline them through to `discr_from_dd` where possible: https://github.com/inducer/grudge/pull/122/commits/ae3a105bfbb6eeda7fbd9209fbecf2283edec474.

I personally like this a lot more

grudge/discretization.py
thomasgibson: Done!

grudge/discretization.py
thomasgibson: Touched up the docs: https://github.com/inducer/grudge/pull/122/commits/6b8458c4cf71e73fa0352fb84cdaea5378bab418.

grudge/discretization.py
thomasgibson: To be honest, I'm starting to like this placeholder business less and less the more I think about it. We need to get this right. `make_discretization_collection` is intended to be the primary user-facing way to create the `DiscretizationCollection` And I can already imagine having an ambiguous placeholder argument causing confusion later on down the line.

I am not a fan, and I think we should avoid them if we can. I would rather just rename `discr_context` back to `mesh`, document it as such, and add an optional kwarg for discretization-related metadata that can be leveraged in the future. Thoughts @inducer @alexfikl ?

grudge/discretization.py
majosm: I haven’t been following this PR closely, so I’m way out of the loop on what exactly is being changed… But I just wanted to ask: what will happen here when we add multiple volume discretizations? Is the `_base_discr` change going to be compatible with that?

grudge/discretization.py
thomasgibson: @majosm This is probably a relevant discussion.

grudge/discretization.py
thomasgibson: The `_base_discr` is specifying the "base"/interpolatory element group on the `mesh`. So one question I have: If you have multiple volume discretizations, will they have different meshes? If so, then _base_discr should probably be a set of "base"/interpolator discretization objects defined on their appropriate volume meshes.

thomasgibson: @majosm When you are at a point where you are trying to couple different discretizations, I would appreciate being involved at some point in the discussion so I can see what's needed in grudge.

majosm: > The `_base_discr` is specifying the "base"/interpolatory element group on the `mesh`. So one question I have: If you have multiple volume discretizations, will they have different meshes? If so, then _base_discr should probably be a set of "base"/interpolator discretization objects defined on their appropriate volume meshes.

I think there will need to be a global mesh (so we have connectivity between the different volumes), but it will get split into sub-meshes for each volume in order to construct the discretizations for each.

thomasgibson: Ah okay! This makes sense. Then it sounds like what we want to make sure is that we pass in some mapping to `make_discretization_collection` (I proposed doing this as some kwarg for metadata) that associates sub-meshes with "base" (often referred to as _the_ volume discretization) discretizations. I would be curious to know your opinion on the matter discussed here: https://github.com/inducer/grudge/pull/122#discussion_r713531929

grudge/discretization.py
thomasgibson: Afraid I don't really know a way around this without resorting to some hacky python. I'd prefer not to make this overly complicated if possible. How about once this PR is merged, we pin an issue describing how the interface has changed?

grudge/discretization.py
inducer: > I'm starting to like this placeholder business less and less the more I think about it. 

Not sure I understand what you're objecting to. Having symbolic names for parts of the domain? Something else? Sorry if I'm being dense.

thomasgibson: Sorry if I wasn't very clear. I was referring to using placeholder arguments in `make_discretization_collection` like `places` or `discr_context` (determined by position alone) which can be pretty much anything (mesh, discretizations, dicts of discretizations). I don't feel comfortable doing that.

thomasgibson: I would rather try to use something along the lines of `**kwargs`, which is a bit more controlled from my perspective.

inducer: Personally, I feel like I've had OK experiences with those types of arguments, if used judiciously. Examples:

- `loopy` functions will often take objects or string representations that parse to them. (e.g. domains, statements)
- `pyopencl.Array`'s constructor can take a context or an array as first argument.

The type of an object can be unambiguously determined, so I don't see a big risk in taking e.g. `mesh` as a shorthand for `{DEFAULT_VOLUME: mesh}`.

On the other hand, I find `**kwargs` parsing kind of unattractive:
- Forget to check for extra arguments
- Assuming you mean we would use `mesh=` or `geometries=` as alternatives to each other, we'd need to add logic to reject calls that have both.
- `**kwargs` have to be the last argument. If that makes the interface kwargs-only, then all calls become kind of wordy, which isn't amazing for high-profile API real estate like the effective discretization collection constructor.

inducer:
Superseded by #236.

Issue #97, opened 2021-05-23
Teach grudge about array containers


MTCam:
[There is a work around for this](https://github.com/illinois-ceesd/mirgecom/blob/d012fbaca0f1c2b0c71c59a200d946fc0c04d9fd/mirgecom/euler.py#L154), but currently this happens when calling `cross_rank_trace_pairs` with a `mirgecom.fluid.ConservedVars` array container.

```
  File "/Users/mtcampbe/CEESD/devel/emirge/grudge/grudge/op.py", line 554, in cross_rank_trace_pairs
    return _cross_rank_trace_pairs_scalar_field(dcoll, ary, tag=tag)
  File "/Users/mtcampbe/CEESD/devel/emirge/grudge/grudge/op.py", line 505, in _cross_rank_trace_pairs_scalar_field
    rbcomms = [_RankBoundaryCommunication(dcoll, remote_rank, vec, tag=tag)
  File "/Users/mtcampbe/CEESD/devel/emirge/grudge/grudge/op.py", line 505, in <listcomp>
    rbcomms = [_RankBoundaryCommunication(dcoll, remote_rank, vec, tag=tag)
  File "/Users/mtcampbe/CEESD/devel/emirge/grudge/grudge/op.py", line 472, in __init__
    local_data = self.array_context.to_numpy(flatten(self.local_dof_array))
  File "/Users/mtcampbe/CEESD/devel/emirge/arraycontext/arraycontext/impl/pyopencl.py", line 292, in to_numpy
    return array.get(queue=self.queue)
AttributeError: 'ConservedVars' object has no attribute 'get'
```

inducer:
Thanks for the report! Ultimately, this will need to take the form of container support in `cross_rank_trace_pairs`.

inducer:
What's missing?

- `cross_rank_trace_pairs`, as originally reported by @MTCam 
- the mass operators
- `op.norm` (perhaps generally the reductions?)
- There are a big bunch of `norm(... .join())`s e.g. [here](https://github.com/illinois-ceesd/mirgecom/blob/eaa9e6cdf82e65a7d90855a4cfe72dc8192acc1a/test/test_euler.py) that could then be deleted.

MTCam:
Is this still necessary? 

```python
    # [Bassi_1997]_ eqn 15 (s = grad_q)    
    grad_cv = make_conserved(dim, q=dg_grad(discr, cv.join(), cv_flux_bnd.join()))
```

[...]

PR #125, opened 2021-06-16
Elementwise min/max and nodal reductions

thomasgibson:
Allows nodal reductions to accept object arrays, and adds a couple new elementwise reductions. This PR also adds targeted unit tests for these reductions.

### DIFF
--- a/grudge/reductions.py
+++ b/grudge/reductions.py
@@ -60,10 +60,7 @@ THE SOFTWARE.
 from numbers import Number
 from functools import reduce
 
-from arraycontext import (
-    ArrayContext,
-    make_loopy_program
-)
+from arraycontext import make_loopy_program
 
 from grudge.discretization import DiscretizationCollection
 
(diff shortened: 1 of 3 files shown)


grudge/reductions.py
```diff
@@ -305,6 +319,92 @@ def elementwise_sum(dcoll: DiscretizationCollection, *args) -> DOFArray:
...
+    return DOFArray(
+        actx,
+        tuple(
+            actx.call_loopy(
+                _map_elementwise_reduction(actx, "max"),
+                operand=vec_i
+            )["result"]
+            for vec_i in vec
+        )
+    )
```
inducer, line 362: Why not share all this code between `sum`/`min`/max`? It's super repetitive, the only difference is the string of the operation name.

thomasgibson: @inducer How's this: https://github.com/inducer/grudge/pull/125/commits/a2b5c49b1292d12ed99fe22d03a6f2df0a027ea6?

grudge/reductions.py
```diff
@@ -156,6 +155,10 @@ def nodal_sum_loc(dcoll: DiscretizationCollection, dd, vec) -> float:
     :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
     :returns: a scalar denoting the rank-local nodal sum.
     """
+    if isinstance(vec, np.ndarray):
```
inducer, line 158: This should really use containers instead of special-casing `numpy` arrays. (below, too)

grudge/reductions.py
```diff
@@ -246,10 +257,35 @@ def integral(dcoll: DiscretizationCollection, dd, vec) -> float:
...
+        dd = dof_desc.DOFDesc("vol", dof_desc.DISCR_TAG_BASE)
+    elif len(args) == 2:
+        dd, vec = args
+    else:
+        raise TypeError("invalid number of arguments")
+
+    dd = dof_desc.as_dofdesc(dd)
+
+    if isinstance(vec, np.ndarray):
+        return obj_array_vectorize(
```
inducer, line 280: This should really use containers instead of special-casing `numpy` arrays.

inducer (commented):
The refactor looks great. Spotted another thing though... sorry!

grudge/reductions.py
thomasgibson: @inducer I totally agree. I was planning to address this exact case in a separate PR (to address https://github.com/inducer/grudge/issues/97 specifically), which will include the reductions and all DG operators. Does that sound okay to you?

inducer: Sure, that's alright by me.

inducer:
Thanks!

PR #130, opened 2021-06-25
Call to_numpy on scalars before doing MPI-reduce on them

kaushikcfd:


inducer:
I think at some point we'll want a distributed-memory abstraction that's lazy-able, but as long as we're calling into MPI, I think this is entirely reasonable.

inducer:
Thanks!

PR #132, opened 2021-06-30
Fix stateful style in TracePair test

inducer:


inducer:
#133 should fix CI fail.

PR #151, opened 2021-07-21
Fix diff property, ugh.

### LINKED ISSUE #150: Diff operator borken
I'm not sure this was as intended...
```python
    @property                                                                                                                      
158    def diff(self):                                                                                                                
159        """A class:`~meshmode.dof_array.DOFArray` or                                                                               
160        :class:`~arraycontext.ArrayContainer` of them representing the                                                             
161        difference (exterior - interior) of the pair values.                                                                       
162        """                                                                                                                        
163        return self.ext - self.ext                                                                                                 
164
```

MTCam:
fixes #150

### DIFF
--- a/grudge/trace_pair.py
+++ b/grudge/trace_pair.py
@@ -160,7 +160,7 @@ class TracePair:
         :class:`~arraycontext.ArrayContainer` of them representing the
         difference (exterior - interior) of the pair values.
         """
-        return self.ext - self.ext
+        return self.ext - self.int
 
 # }}}
 
--- a/grudge/trace_pair.py
+++ b/grudge/trace_pair.py
@@ -86,6 +86,7 @@ class TracePair:
     .. autoattribute:: int
     .. autoattribute:: ext
     .. autoattribute:: avg
+    .. autoattribute:: diff
 
     .. automethod:: __getattr__
     .. automethod:: __getitem__
--- a/test/test_trace_pair.py
+++ b/test/test_trace_pair.py
@@ -62,6 +62,7 @@ def test_trace_pair(actx_factory):
     exterior = rand()
     tpair = TracePair("vol", interior=interior, exterior=exterior)
 
+    import grudge.op as op
     assert op.norm(dcoll, tpair.avg - 0.5*(exterior + interior), np.inf) == 0
     assert op.norm(dcoll, tpair.diff - (exterior - interior), np.inf) == 0
     assert op.norm(dcoll, tpair.int - interior, np.inf) == 0
--- a/test/test_trace_pair.py
+++ b/test/test_trace_pair.py
@@ -62,7 +62,7 @@ def test_trace_pair(actx_factory):
     exterior = rand()
     tpair = TracePair("vol", interior=interior, exterior=exterior)
 
-    assert tpair.avg == 0.5*(interior + exterior)
-    assert tpair.diff == (exterior - interior)
-    assert tpair.int == interior
-    assert tpair.ext == exterior
+    assert op.norm(dcoll, tpair.avg - 0.5*(exterior + interior), np.inf) == 0
+    assert op.norm(dcoll, tpair.diff - (exterior - interior), np.inf) == 0
+    assert op.norm(dcoll, tpair.int - interior, np.inf) == 0
+    assert op.norm(dcoll, tpair.ext - exterior, np.inf) == 0


matthiasdiener:
Could also add this to the class doc maybe

thomasgibson (approved):
Thanks for catching this @MTCam

MTCam:
> Could also add this method to the class doc maybe

Implementation is already more doc than code!
```python
        """A class:`~meshmode.dof_array.DOFArray` or                                                                               
160        :class:`~arraycontext.ArrayContainer` of them representing the                                                             
161        difference (exterior - interior) of the pair values.                                                                       
162        """                                                                                                                        
163
```

matthiasdiener:
I mean, just adding `.. autoattribute:: diff` above for completeness.

MTCam:
> Thanks for catching this @MTCam

I *didn't*.  *MIRGE-Com* production testing in CI caught this, and only because isolator (which was just added a few days ago!) runs enough steps to catch the solution diverging from that which is expected.

matthiasdiener (approved):

MTCam:
> I mean, just adding `.. autoattribute:: diff` above for completeness.

Ugh, sorry yeah (b8f3720)

alexfikl:
There's a test for this in [test_trace_pair.py](https://github.com/inducer/grudge/blob/a1e96b35755a16f2f013a265a40e150a5c246fce/test/test_trace_pair.py#L66). How come that's passing?

MTCam:
> There's a test for this in [test_trace_pair.py](https://github.com/inducer/grudge/blob/a1e96b35755a16f2f013a265a40e150a5c246fce/test/test_trace_pair.py#L66). How come that's passing?

I can only guess because the two calls to `rand()` are returning the same random number?  🤔

matthiasdiener:
> There's a test for this in [test_trace_pair.py](https://github.com/inducer/grudge/blob/a1e96b35755a16f2f013a265a40e150a5c246fce/test/test_trace_pair.py#L66). How come that's passing?

Is the assert testing element equality?

thomasgibson:
> > There's a test for this in [test_trace_pair.py](https://github.com/inducer/grudge/blob/a1e96b35755a16f2f013a265a40e150a5c246fce/test/test_trace_pair.py#L66). How come that's passing?
> 
> Is the assert testing element equality?

I actually don't know why it didn't scream before; I just tested locally and this definitely should have failed the assert...

alexfikl:
> I actually don't know why it didn't scream before; I just tested locally and this definitely should have failed the assert...

It's likely because `DOFArray` implements `__len__` so it gets a "truthy" value by default
https://docs.python.org/3/library/stdtypes.html#truth-value-testing
It should add a `__bool__` that raises in a similar way that doing `bool(my_numpy_array)` does.

MTCam:
> > I actually don't know why it didn't scream before; I just tested locally and this definitely should have failed the assert...
> 
> It's likely because `DOFArray` implements `__len__` so it gets a "truthy" value by default
> [docs.python.org/3/library/stdtypes.html#truth-value-testing](https://docs.python.org/3/library/stdtypes.html#truth-value-testing)
> It should add a `__bool__` that raises in a similar way that doing `bool(my_numpy_array)` does.

I wonder how many of our other tests will jump up-and-down screaming if this change is made.

thomasgibson:
@alexfikl I see. Thanks for the explaination. Yeah we need to be careful about documenting this at some point. @MTCam Can you add this
```python
    assert op.norm(dcoll, tpair.avg - 0.5*(exterior + interior), np.inf) == 0
    assert op.norm(dcoll, tpair.diff - (exterior - interior), np.inf) == 0
    assert op.norm(dcoll, tpair.int - interior, np.inf) == 0
    assert op.norm(dcoll, tpair.ext - exterior, np.inf) == 0
```
in `test_trace_pair.py` (specifically in lines 66-69)? At least in this way we are _numerically_ comparing them (which I think we want to be doing here). @matthiasdiener @alexfikl any thoughts?

MTCam:
> @alexfikl I see. Thanks for the explaination. Yeah we need to be careful about documenting this at some point. @MTCam Can you add this
> 
> ```python
>     assert op.norm(dcoll, tpair.avg - 0.5*(exterior + interior), np.inf) == 0
>     assert op.norm(dcoll, tpair.diff - (exterior - interior), np.inf) == 0
>     assert op.norm(dcoll, tpair.int - interior, np.inf) == 0
>     assert op.norm(dcoll, tpair.ext - exterior, np.inf) == 0
> ```
> 
> in `test_trace_pair.py` (specifically in lines 66-69)? At least in this way we are _numerically_ comparing them (which I think we want to be doing here). @matthiasdiener @alexfikl any thoughts?

(35b9676)

PR #152, opened 2021-08-10
Adds hello-grudge example on the landing page

kaushikcfd:


doc/index.rst
```diff
@@ -1,6 +1,26 @@
...
 ==================================
 
+Here's an example to solve the PDE
+
+.. math::
+    \frac{\partial u}{\partial t} + 2\pi \frac{\partial u}{\partial x} = 0
+.. math::
+    u(x, 0) = \sin(x)
+.. math::
+    u(0, t) = -\sin(2\pi t)
```
alexfikl, line 11:
```suggestion
.. math::
    \begin{cases}
    u_t + 2\pi u_x = 0, \\
    u(0, t) = -\sin(2\pi t), \\
    u(x, 0) = \sin(x),
    \end{cases}
```
?

examples/hello-grudge.py
```diff
@@ -0,0 +1,120 @@
+import numpy as np
+import pyopencl as cl
+from grudge.discretization import DiscretizationCollection
+import grudge.op as op
+from meshmode.mesh.generation import generate_box_mesh
+from meshmode.array_context import PyOpenCLArrayContext
+from arraycontext import freeze, thaw
+from grudge.dof_desc import DTAG_BOUNDARY
+from meshmode.discretization.connection import FACE_RESTR_INTERIOR
```
alexfikl, line 9:
```suggestion
from grudge.dof_desc import DTAG_BOUNDARY, FACE_RESTR_INTERIOR
```
It doesn't seem documented, but `FACE_RESTR_INTERIOR` is exported in `dof_desc` too.

examples/hello-grudge.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+ctx = cl.create_some_context()
+queue = cl.CommandQueue(ctx)
+actx = PyOpenCLArrayContext(queue)
+
+
```
alexfikl, line 16:
```suggestion
```

examples/hello-grudge.py
```diff
@@ -0,0 +1,120 @@
...
+nel = 10
+coords = np.linspace(0, 2*np.pi, nel)
+mesh = generate_box_mesh((coords,),
+                         boundary_tag_to_face={"left": ["-x"],
+                                               "right": ["+x"]})
```
alexfikl, line 21: Why does an advection equation need a "right" boundary? It seems to just be upwinding.

examples/hello-grudge.py
```diff
@@ -0,0 +1,120 @@
...
+
+x_vol = thaw(dcoll.nodes(), actx)
+x_bndry = thaw(dcoll.discr_from_dd(left_bndry).nodes(), actx)
+
+uh = initial_condition(x_vol)
+
+dt = 0.001
+t = 0
+t_final = 0.5
+
```
alexfikl, line 58:
```suggestion
```

examples/hello-grudge.py
```diff
@@ -0,0 +1,120 @@
...
+    return actx.np.sin(x[0])
+
+
+def left_boundary_condition(x, t):
+    return actx.np.sin(x[0] - 2 * np.pi * t)
+
+
+def flux(dcoll, u_tpair):
+    dd = u_tpair.dd
+    velocity = np.array([2 * np.pi])
```
alexfikl, line 37: Set the velocity field as a global? It seems to be used in several places.

alexfikl (approved):
Left a few nitpicks, but overall big :+1:!

examples/hello-grudge.py
kaushikcfd: For the right boundary we calculate flux similar to the interior faces. Need to include here as the interior_faces discretization (righfully) doesn't include the rightmost face of the domain.

examples/hello-grudge.py
kaushikcfd: I specifically chose to avoid this to keep the hello-world example as "flat" as possible. My main concern with the other examples in `grudge/examples` was the indirection in control was making it difficult to follow along what was happening.

alexfikl: I agree that it's easy to read as is, so I'm fine with it.

examples/hello-grudge.py
alexfikl: Ah, right, you still have an outgoing flux that needs to be handled.

Makes sense!

inducer:
Any idea why the CI is refusing to run?

inducer:
(Also, thanks for adding this, and thanks @alexfikl for taking a look!)

kaushikcfd:
Yesterday there was a GithubCI outage, probably it's still stuck because of that. https://www.githubstatus.com/incidents/lc8s9rtyjmgq

examples/hello-grudge.py
```diff
@@ -0,0 +1,117 @@
...
+
+uh = initial_condition(x_vol)
+
+dt = 0.001
+t = 0
+t_final = 0.5
+
+# timestepper loop
+while t < t_final:
+    uh = thaw(freeze(uh, actx), actx)
```
inducer, line 59: Why is this here? (I'm guessing something to do with lazy? If so, comment.

examples/hello-grudge.py
```diff
@@ -0,0 +1,117 @@
...
+import pyopencl as cl
+from grudge.discretization import DiscretizationCollection
+import grudge.op as op
+from meshmode.mesh.generation import generate_box_mesh
+from meshmode.array_context import PyOpenCLArrayContext
+from arraycontext import freeze, thaw
+from grudge.dof_desc import DTAG_BOUNDARY, FACE_RESTR_INTERIOR
+
+
+ctx = cl.create_some_context()
```
inducer, line 11: Add a comment describing what this is. (I'm guessing you'd prefer to not move it under `advection`, right?)

inducer (commented):
Thanks! Just two readability concerns, otherwise good to go.

examples/hello-grudge.py
kaushikcfd: Since this was really only needed for the lazy mode, avoided it to keep things simple. (Thanks!)

examples/hello-grudge.py
kaushikcfd: Added a comment [here](https://github.com/inducer/grudge/pull/152/commits/76e506d21dc1e14b25b3748658fffa0747cfe103#diff-f740fc40d4a29a0d4743d76a19fef4be2716f7bb0a647cb73466382c225193afR11). How does it look?

inducer: Oh, sorry! I meant maybe more a (short!) comment docstring describing what this file is, for folks who don't arrive via the docs. (Maybe move the descriptive blurb from the docs?)

kaushikcfd: Oh :). Done in https://github.com/inducer/grudge/pull/152/commits/464e5e2cea5e6c192551cdf697b29be113ebbbae#diff-f740fc40d4a29a0d4743d76a19fef4be2716f7bb0a647cb73466382c225193afR1-R10.

inducer:
LGTM! Thanks for working on this.

PR #154, opened 2021-08-17
Array container support and batched (eager) communication

thomasgibson:
This PR modifies existing grudge routines to work on arbitrary array containers and batches eager MPI communication of containers.

~Requires: https://github.com/inducer/arraycontext/pull/91~

### DIFF
--- a/examples/wave/wave-op-mpi.py
+++ b/examples/wave/wave-op-mpi.py
@@ -98,8 +98,9 @@ def wave_operator(dcoll, c, w):
     u = w.u
     v = w.v
 
-    dir_u = op.project(dcoll, "vol", BTAG_ALL, u)
-    dir_v = op.project(dcoll, "vol", BTAG_ALL, v)
+    dir_w = op.project(dcoll, "vol", BTAG_ALL, w)
+    dir_u = dir_w.u
+    dir_v = dir_w.v
     dir_bval = WaveState(u=dir_u, v=dir_v)
     dir_bc = WaveState(u=-dir_u, v=dir_v)
 
--- a/examples/wave/wave-op-mpi.py
+++ b/examples/wave/wave-op-mpi.py
@@ -67,7 +67,6 @@ class WaveState:
 
     def __post_init__(self):
         assert isinstance(self.v, np.ndarray) and self.v.dtype.char == "O"
-        assert all(vi.array_context is self.u.array_context for vi in self.v)
 
     @property
     def array_context(self):
--- a/examples/wave/wave-op-mpi.py
+++ b/examples/wave/wave-op-mpi.py
@@ -67,6 +67,7 @@ class WaveState:
 
     def __post_init__(self):
         assert isinstance(self.v, np.ndarray) and self.v.dtype.char == "O"
+        assert all(vi.array_context is self.u.array_context for vi in self.v)
 
     @property
     def array_context(self):
@@ -80,15 +81,15 @@ def wave_flux(dcoll, c, w_tpair):
     normal = thaw(dcoll.normal(w_tpair.dd), u.int.array_context)
 
     flux_weak = WaveState(
-        u=np.dot(v.avg, normal),
-        v=normal*u.avg
+        u=v.avg @ normal,
+        v=u.avg * normal
     )
 
     # upwind
-    v_jump = np.dot(normal, v.ext-v.int)
+    v_jump = v.diff @ normal
     flux_weak += WaveState(
-        u=0.5*(u.ext-u.int),
-        v=0.5*normal*v_jump,
+        u=0.5 * u.diff,
+        v=0.5 * v_jump * normal,
     )
 
     return op.project(dcoll, w_tpair.dd, "all_faces", c*flux_weak)
--- a/examples/wave/wave-op-mpi.py
+++ b/examples/wave/wave-op-mpi.py
@@ -31,11 +31,18 @@ import numpy.linalg as la  # noqa
 import pyopencl as cl
 import pyopencl.tools as cl_tools
 
-from arraycontext import thaw
+from arraycontext import (
+    thaw,
+    with_container_arithmetic,
+    dataclass_array_container
+)
 from grudge.array_context import PyOpenCLArrayContext
 
-from pytools.obj_array import flat_obj_array
+from dataclasses import dataclass
 
+from pytools.obj_array import flat_obj_array, make_obj_array
+
+from meshmode.dof_array import DOFArray
 from meshmode.mesh import BTAG_ALL, BTAG_NONE  # noqa
 
 from grudge.discretization import DiscretizationCollection
@@ -51,42 +58,57 @@ from mpi4py import MPI
 
 # {{{ wave equation bits
 
+@with_container_arithmetic(bcast_obj_array=True, rel_comparison=True)
+@dataclass_array_container
+@dataclass(frozen=True)
+class WaveState:
+    u: DOFArray
+    v: np.ndarray  # [object array]
+
+    def __post_init__(self):
+        assert isinstance(self.v, np.ndarray) and self.v.dtype.char == "O"
+
+    @property
+    def array_context(self):
+        return self.u.array_context
+
+
 def wave_flux(dcoll, c, w_tpair):
-    u = w_tpair[0]
-    v = w_tpair[1:]
+    u = w_tpair.u
+    v = w_tpair.v
 
     normal = thaw(dcoll.normal(w_tpair.dd), u.int.array_context)
 
-    flux_weak = flat_obj_array(
-            np.dot(v.avg, normal),
-            normal*u.avg,
-            )
+    flux_weak = WaveState(
+        u=np.dot(v.avg, normal),
+        v=normal*u.avg
+    )
 
     # upwind
     v_jump = np.dot(normal, v.ext-v.int)
-    flux_weak += flat_obj_array(
-            0.5*(u.ext-u.int),
-            0.5*normal*v_jump,
-            )
+    flux_weak += WaveState(
+        u=0.5*(u.ext-u.int),
+        v=0.5*normal*v_jump,
+    )
 
     return op.project(dcoll, w_tpair.dd, "all_faces", c*flux_weak)
 
 
 def wave_operator(dcoll, c, w):
-    u = w[0]
-    v = w[1:]
+    u = w.u
+    v = w.v
 
     dir_u = op.project(dcoll, "vol", BTAG_ALL, u)
     dir_v = op.project(dcoll, "vol", BTAG_ALL, v)
-    dir_bval = flat_obj_array(dir_u, dir_v)
-    dir_bc = flat_obj_array(-dir_u, dir_v)
+    dir_bval = WaveState(u=dir_u, v=dir_v)
+    dir_bc = WaveState(u=-dir_u, v=dir_v)
 
     return (
         op.inverse_mass(
             dcoll,
-            flat_obj_array(
-                -c*op.weak_local_div(dcoll, v),
-                -c*op.weak_local_grad(dcoll, u)
+            WaveState(
+                u=-c*op.weak_local_div(dcoll, v),
+                v=-c*op.weak_local_grad(dcoll, u)
             )
             + op.face_mass(
                 dcoll,
@@ -179,10 +201,10 @@ def main(ctx_factory, dim=2, order=3, visualize=False):
     dcoll = DiscretizationCollection(actx, local_mesh, order=order,
                     mpi_communicator=comm)
 
-    fields = flat_obj_array(
-            bump(actx, dcoll),
-            [dcoll.zeros(actx) for i in range(dcoll.dim)]
-            )
+    fields = WaveState(
+        u=bump(actx, dcoll),
+        v=make_obj_array([dcoll.zeros(actx) for i in range(dcoll.dim)])
+    )
 
     c = 1
     dt = 0.45 * estimate_rk4_timestep(actx, dcoll, c)
@@ -201,12 +223,12 @@ def main(ctx_factory, dim=2, order=3, visualize=False):
     while t < t_final:
         fields = rk4_step(fields, t, dt, rhs)
 
-        l2norm = op.norm(dcoll, fields[0], 2)
+        l2norm = op.norm(dcoll, fields.u, 2)
 
         if istep % 10 == 0:
-            linfnorm = op.norm(dcoll, fields[0], np.inf)
-            nodalmax = op.nodal_max(dcoll, "vol", fields[0])
-            nodalmin = op.nodal_min(dcoll, "vol", fields[0])
+            linfnorm = op.norm(dcoll, fields.u, np.inf)
+            nodalmax = op.nodal_max(dcoll, "vol", fields.u)
+            nodalmin = op.nodal_min(dcoll, "vol", fields.u)
             if comm.rank == 0:
                 logger.info(f"step: {istep} t: {t} "
                             f"L2: {l2norm} "
@@ -218,8 +240,8 @@ def main(ctx_factory, dim=2, order=3, visualize=False):
                     comm,
                     f"fld-wave-eager-mpi-{{rank:03d}}-{istep:04d}.vtu",
                     [
-                        ("u", fields[0]),
-                        ("v", fields[1:]),
+                        ("u", fields.u),
+                        ("v", fields.v),
                     ]
                 )
 
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -54,7 +54,7 @@ from arraycontext import ArrayContext, map_array_container
 
 from functools import partial
 
-from meshmode.dof_array import DOFArray
+from meshmode.dof_array import DOFArray, rec_map_dof_array_container
 from meshmode.transform_metadata import FirstAxisIsElementsTag
 
 from grudge.discretization import DiscretizationCollection
@@ -520,7 +520,7 @@ def reference_mass_matrix(actx: ArrayContext, out_element_group, in_element_grou
 def _apply_mass_operator(
         dcoll: DiscretizationCollection, dd_out, dd_in, vec):
     if not isinstance(vec, DOFArray):
-        return map_array_container(
+        return rec_map_dof_array_container(
             partial(_apply_mass_operator, dcoll, dd_out, dd_in), vec
         )
 
@@ -617,7 +617,7 @@ def reference_inverse_mass_matrix(actx: ArrayContext, element_group):
 def _apply_inverse_mass_operator(
         dcoll: DiscretizationCollection, dd_out, dd_in, vec):
     if not isinstance(vec, DOFArray):
-        return map_array_container(
+        return rec_map_dof_array_container(
             partial(_apply_inverse_mass_operator, dcoll, dd_out, dd_in), vec
         )
 
@@ -788,7 +788,7 @@ def reference_face_mass_matrix(
 
 def _apply_face_mass_operator(dcoll: DiscretizationCollection, dd, vec):
     if not isinstance(vec, DOFArray):
-        return map_array_container(
+        return rec_map_dof_array_container(
             partial(_apply_face_mass_operator, dcoll, dd), vec
         )
 
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -198,6 +198,15 @@ def local_grad(dcoll: DiscretizationCollection, vec) -> ArrayOrContainerT:
         :class:`~arraycontext.container.ArrayContainer`\ of object arrays.
     """
     if not isinstance(vec, DOFArray):
+        # FIXME: Occasionally, data structures coming from *mirgecom* will
+        # contain empty object arrays as placeholders for fields.
+        # For example, species mass fractions is an empty object array when
+        # running in a single-species configuration.
+        # This hack here ensures that these empty arrays, at the very least,
+        # have their shape updated when applying the gradient operator
+        if (isinstance(vec, np.ndarray) and vec.dtype == object
+                and vec.size == 0):
+            vec = vec.reshape((0, dcoll.dim))
         return map_array_container(partial(local_grad, dcoll), vec)
 
     from grudge.geometry import inverse_surface_metric_derivative_mat
@@ -245,9 +254,14 @@ def local_d_dx(
 
 
 def _div_helper(diff_func, vecs):
-    if not isinstance(vecs, np.ndarray):
+    if not (isinstance(vecs, np.ndarray) and vecs.dtype == object):
         raise TypeError("argument must be an object array")
-    assert vecs.dtype == object
+
+    if vecs.size == 0:
+        # FIXME: Need to handle the special case where *mirgecom* passes in
+        # an empty object array to denote that there are no multiple species
+        # in the state.
+        return np.empty(shape=(0,), dtype=vecs.dtype)
 
     return sum(diff_func(i, vec_i) for i, vec_i in enumerate(vecs))
 
@@ -268,7 +282,7 @@ def local_div(dcoll: DiscretizationCollection, vecs) -> ArrayOrContainerT:
     :returns: a :class:`~meshmode.dof_array.DOFArray` or an
         :class:`~arraycontext.container.ArrayContainer` of them.
     """
-    if not (isinstance(vecs, np.ndarray) and vecs.dtype == "O"):
+    if not (isinstance(vecs, np.ndarray) and vecs.dtype == object):
         return map_array_container(partial(local_div, dcoll), vecs)
 
     return _div_helper(lambda i, subvec: local_d_dx(dcoll, i, subvec), vecs)
@@ -354,6 +368,15 @@ def weak_local_grad(dcoll: DiscretizationCollection, *args) -> ArrayOrContainerT
         raise TypeError("invalid number of arguments")
 
     if not isinstance(vec, DOFArray):
+        # FIXME: Occasionally, data structures coming from *mirgecom* will
+        # contain empty object arrays as placeholders for fields.
+        # For example, species mass fractions is an empty object array when
+        # running in a single-species configuration.
+        # This hack here ensures that these empty arrays, at the very least,
+        # have their shape updated when applying the gradient operator
+        if (isinstance(vec, np.ndarray) and vec.dtype == object
+                and vec.size == 0):
+            vec = vec.reshape((0, dcoll.dim))
         return map_array_container(partial(weak_local_grad, dcoll, dd_in), vec)
 
     from grudge.geometry import inverse_surface_metric_derivative_mat
@@ -468,7 +491,7 @@ def weak_local_div(dcoll: DiscretizationCollection, *args) -> ArrayOrContainerT:
     else:
         raise TypeError("invalid number of arguments")
 
-    if not (isinstance(vecs, np.ndarray) and vecs.dtype == "O"):
+    if not (isinstance(vecs, np.ndarray) and vecs.dtype == object):
         return map_array_container(partial(weak_local_div, dcoll, dd), vecs)
 
     return _div_helper(
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -54,7 +54,7 @@ from arraycontext import ArrayContext, map_array_container
 
 from functools import partial
 
-from meshmode.dof_array import DOFArray, rec_map_dof_array_container
+from meshmode.dof_array import DOFArray
 from meshmode.transform_metadata import FirstAxisIsElementsTag
 
 from grudge.discretization import DiscretizationCollection
@@ -520,7 +520,7 @@ def reference_mass_matrix(actx: ArrayContext, out_element_group, in_element_grou
 def _apply_mass_operator(
         dcoll: DiscretizationCollection, dd_out, dd_in, vec):
     if not isinstance(vec, DOFArray):
-        return rec_map_dof_array_container(
+        return map_array_container(
             partial(_apply_mass_operator, dcoll, dd_out, dd_in), vec
         )
 
@@ -617,7 +617,7 @@ def reference_inverse_mass_matrix(actx: ArrayContext, element_group):
 def _apply_inverse_mass_operator(
         dcoll: DiscretizationCollection, dd_out, dd_in, vec):
     if not isinstance(vec, DOFArray):
-        return rec_map_dof_array_container(
+        return map_array_container(
             partial(_apply_inverse_mass_operator, dcoll, dd_out, dd_in), vec
         )
 
@@ -788,7 +788,7 @@ def reference_face_mass_matrix(
 
 def _apply_face_mass_operator(dcoll: DiscretizationCollection, dd, vec):
     if not isinstance(vec, DOFArray):
-        return rec_map_dof_array_container(
+        return map_array_container(
             partial(_apply_face_mass_operator, dcoll, dd), vec
         )
 
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -51,6 +51,7 @@ THE SOFTWARE.
 
 
 from arraycontext import ArrayContext, map_array_container
+from arraycontext.container import ArrayOrContainerT
 
 from functools import partial
 
@@ -189,12 +190,13 @@ def local_grad(
         \nabla|_E f = \left(
             \partial_x|_E f, \partial_y|_E f, \partial_z|_E f \right)
 
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     :arg nested: return nested object arrays instead of a single multidimensional
         array if *vec* is non-scalar.
     :returns: an object array (possibly nested) of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+        :class:`~meshmode.dof_array.DOFArray`\ s or
+        :class:`~arraycontext.container.ArrayContainer`\ s.
     """
     if isinstance(vec, np.ndarray):
         grad = obj_array_vectorize(
@@ -216,7 +218,8 @@ def local_grad(
             metric_in_matvec=False)
 
 
-def local_d_dx(dcoll: DiscretizationCollection, xyz_axis, vec):
+def local_d_dx(
+        dcoll: DiscretizationCollection, xyz_axis, vec) -> ArrayOrContainerT:
     r"""Return the element-local derivative along axis *xyz_axis* of a
     function :math:`f` represented by *vec*:
 
@@ -226,8 +229,10 @@ def local_d_dx(dcoll: DiscretizationCollection, xyz_axis, vec):
 
     :arg xyz_axis: an integer indicating the axis along which the derivative
         is taken.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
-    :returns: a :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     """
     discr = dcoll.discr_from_dd(dof_desc.DD_VOLUME)
     actx = vec.array_context
@@ -265,7 +270,7 @@ def _div_helper(dcoll: DiscretizationCollection, diff_func, vecs):
         return result
 
 
-def local_div(dcoll: DiscretizationCollection, vecs):
+def local_div(dcoll: DiscretizationCollection, vecs) -> ArrayOrContainerT:
     r"""Return the element-local divergence of the vector function
     :math:`\mathbf{f}` represented by *vecs*:
 
@@ -274,10 +279,12 @@ def local_div(dcoll: DiscretizationCollection, vecs):
         \nabla|_E \cdot \mathbf{f} = \sum_{i=1}^d \partial_{x_i}|_E \mathbf{f}_i
 
     :arg vec: an object array of
-        a :class:`~meshmode.dof_array.DOFArray`\ s,
+        :class:`~meshmode.dof_array.DOFArray`\ s or
+        :class:`~arraycontext.container.ArrayContainer` objects,
         where the last axis of the array must have length
         matching the volume dimension.
-    :returns: a :class:`~meshmode.dof_array.DOFArray`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     """
 
     return _div_helper(dcoll,
@@ -334,7 +341,8 @@ def _reference_stiffness_transpose_matrix(
                                            in_element_group)
 
 
-def weak_local_grad(dcoll: DiscretizationCollection, *args, nested=False):
+def weak_local_grad(
+        dcoll: DiscretizationCollection, *args, nested=False) -> np.ndarray:
     r"""Return the element-local weak gradient of the volume function
     represented by *vec*.
 
@@ -348,12 +356,13 @@ def weak_local_grad(dcoll: DiscretizationCollection, *args, nested=False):
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
         Defaults to the base volume discretization if not provided.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     :arg nested: return nested object arrays instead of a single multidimensional
         array if *vec* is non-scalar
     :returns: an object array (possibly nested) of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+        :class:`~meshmode.dof_array.DOFArray`\ s or
+        :class:`~arraycontext.container.ArrayContainer`\ s.
     """
     if len(args) == 1:
         vec, = args
@@ -386,7 +395,7 @@ def weak_local_grad(dcoll: DiscretizationCollection, *args, nested=False):
             metric_in_matvec=True)
 
 
-def weak_local_d_dx(dcoll: DiscretizationCollection, *args):
+def weak_local_d_dx(dcoll: DiscretizationCollection, *args) -> ArrayOrContainerT:
     r"""Return the element-local weak derivative along axis *xyz_axis* of the
     volume function represented by *vec*.
 
@@ -409,8 +418,10 @@ def weak_local_d_dx(dcoll: DiscretizationCollection, *args):
 
     :arg xyz_axis: an integer indicating the axis along which the derivative
         is taken
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
-    :returns: a :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     """
     if len(args) == 2:
         xyz_axis, vec = args
@@ -436,7 +447,7 @@ def weak_local_d_dx(dcoll: DiscretizationCollection, *args):
             metric_in_matvec=True)
 
 
-def weak_local_div(dcoll: DiscretizationCollection, *args):
+def weak_local_div(dcoll: DiscretizationCollection, *args) -> ArrayOrContainerT:
     r"""Return the element-local weak divergence of the vector volume function
     represented by *vecs*.
 
@@ -457,11 +468,13 @@ def weak_local_div(dcoll: DiscretizationCollection, *args):
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
         Defaults to the base volume discretization if not provided.
-    :arg vec: a object array of
-        a :class:`~meshmode.dof_array.DOFArray`\ s,
+    :arg vec: an object array of
+        :class:`~meshmode.dof_array.DOFArray`\s or
+        :class:`~arraycontext.container.ArrayContainer` objects,
         where the last axis of the array must have length
         matching the volume dimension.
-    :returns: a :class:`~meshmode.dof_array.DOFArray`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     """
     if len(args) == 1:
         vecs, = args
@@ -552,10 +565,10 @@ def _apply_mass_operator(
     )
 
 
-def mass(dcoll: DiscretizationCollection, *args):
+def mass(dcoll: DiscretizationCollection, *args) -> ArrayOrContainerT:
     r"""Return the action of the DG mass matrix on a vector (or vectors)
     of :class:`~meshmode.dof_array.DOFArray`\ s, *vec*. In the case of
-    *vec* being an object array of :class:`~meshmode.dof_array.DOFArray`\ s,
+    *vec* being an :class:`~arraycontext.container.ArrayContainer`,
     the mass operator is applied in the Kronecker sense (component-wise).
 
     May be called with ``(vec)`` or ``(dd, vec)``.
@@ -572,11 +585,10 @@ def mass(dcoll: DiscretizationCollection, *args):
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
         Defaults to the base volume discretization if not provided.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
-    :returns: a :class:`~meshmode.dof_array.DOFArray` denoting the
-        application of the mass matrix, or an object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     """
 
     if len(args) == 1:
@@ -653,12 +665,11 @@ def _apply_inverse_mass_operator(
     return DOFArray(actx, data=tuple(group_data))
 
 
-def inverse_mass(dcoll: DiscretizationCollection, vec):
+def inverse_mass(dcoll: DiscretizationCollection, vec) -> ArrayOrContainerT:
     r"""Return the action of the DG mass matrix inverse on a vector
     (or vectors) of :class:`~meshmode.dof_array.DOFArray`\ s, *vec*.
-    In the case of *vec* being an object array of
-    :class:`~meshmode.dof_array.DOFArray`\ s, the inverse mass operator is
-    applied in the Kronecker sense (component-wise).
+    In the case of *vec* being an :class:`~arraycontext.container.ArrayContainer`,
+    the inverse mass operator is applied in the Kronecker sense (component-wise).
 
     For affine elements :math:`E`, the element-wise mass inverse
     is computed directly as the inverse of the (physical) mass matrix:
@@ -684,11 +695,10 @@ def inverse_mass(dcoll: DiscretizationCollection, vec):
     where :math:`\widehat{\mathbf{M}}` is the reference mass matrix on
     :math:`\widehat{E}`.
 
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
-    :returns: a :class:`~meshmode.dof_array.DOFArray` denoting the
-        application of the inverse mass matrix, or an object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     """
 
     return _apply_inverse_mass_operator(
@@ -831,11 +841,11 @@ def _apply_face_mass_operator(dcoll: DiscretizationCollection, dd, vec):
     )
 
 
-def face_mass(dcoll: DiscretizationCollection, *args):
+def face_mass(dcoll: DiscretizationCollection, *args) -> ArrayOrContainerT:
     r"""Return the action of the DG face mass matrix on a vector (or vectors)
     of :class:`~meshmode.dof_array.DOFArray`\ s, *vec*. In the case of
-    *vec* being an object array of :class:`~meshmode.dof_array.DOFArray`\ s,
-    the mass operator is applied in the Kronecker sense (component-wise).
+    *vec* being an arbitrary :class:`~arraycontext.container.ArrayContainer`,
+    the face mass operator is applied in the Kronecker sense (component-wise).
 
     May be called with ``(vec)`` or ``(dd, vec)``.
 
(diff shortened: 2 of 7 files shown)


grudge/trace_pair.py
```diff
@@ -276,57 +284,109 @@ class _RankBoundaryCommunication:
...
+                )
+                offset += _ndofs
+            # Store the unflattened, de-numpy-fied remote data
+            remote_container_data[key] = remote_data
+
+        # Re-containerize the remote data
+        remote_array_ctr = deserialize_container(
+            self.local_array_ctr,
+            remote_container_data.items()
         )
```
thomasgibson, line 378: @inducer I'm very unhappy with how this turned out, but it seems to work. However, I think it is too inefficient (performance seems to get worse the more ranks are communicating). The challenge here was trying to "unconvert" an array container whose components have been flattened and concatenated into a single numpy array to be communicated (see https://github.com/inducer/grudge/pull/154/files#diff-5fa6e6b713d37028c16f1bbc5a6d0b4547ecba0f86c8a52eb30e1ba1dc2614e4R322-R330).

This needs to be revisited before I am comfortable merging this. I would appreciate some insight on this.

grudge/trace_pair.py
```diff
@@ -276,57 +284,109 @@ class _RankBoundaryCommunication:
...
+                partial(flatten_to_numpy, self.array_context), value
+            )
+            container_data[key] = flattened_values
+            # NOTE: Addresses object arrays inside array containers. Need to special
+            # case them here to make sure we can convert it into a flat numpy array
+            # that mpi4py can actually send/recieve
+            if (isinstance(flattened_values, np.ndarray)
+                    and flattened_values.dtype == "O"):
+                flattened_values = np.concatenate(flattened_values)
+            flattened_data.append(flattened_values)
```
thomasgibson, line 317: The reason for this `if`-statement here is that I'm trying to avoid _all_ sub array containers (in particular, object numpy arrays). `meshmode.dof_array.flatten_to_numpy` (and similar functions) appear to only be implemented with flattening group-wise in mind. As a result, I have to add these special cases (same in `finish()` when converting _back_ to an array container).

grudge/op.py
```diff
@@ -478,11 +481,9 @@ def get_ref_mass_mat(out_grp, in_grp):
 
 def _apply_mass_operator(
         dcoll: DiscretizationCollection, dd_out, dd_in, vec):
-    if isinstance(vec, np.ndarray):
-        return obj_array_vectorize(
-            lambda vi: _apply_mass_operator(dcoll,
-                                            dd_out,
-                                            dd_in, vi), vec
+    if not isinstance(vec, DOFArray):
+        return rec_map_dof_array_container(
```
alexfikl, line 485:
```suggestion
        return map_array_container(
```
There's a lot of traversal functions around these days :\\. The `rec_` versions will basically recurse through the array container themselves until they hit a non-array container. The `dof_array_` versions recurse until they hit a `DOFArray` **or** a non-array container.

For this code, it would be better to
* use the non-recursive `map_dof_array_container`, since it's calling the function again itself, i.e. implementing the recursion right here.
* use the standard `map_array_container`, since it's checking for `DOFArray` itself.
* \+ `map_array_container` is probably the fastest of the bunch, since it has to do the least amount of work.

This should apply to the other changes below too.

grudge/op.py
```diff
@@ -576,11 +577,9 @@ def get_ref_inv_mass_mat(grp):
 
 def _apply_inverse_mass_operator(
         dcoll: DiscretizationCollection, dd_out, dd_in, vec):
-    if isinstance(vec, np.ndarray):
-        return obj_array_vectorize(
-            lambda vi: _apply_inverse_mass_operator(dcoll,
-                                                    dd_out,
-                                                    dd_in, vi), vec
+    if not isinstance(vec, DOFArray):
+        return rec_map_dof_array_container(
```
alexfikl, line 581:
```suggestion
        return map_array_container(
```

grudge/op.py
```diff
@@ -748,9 +747,9 @@ def get_ref_face_mass_mat(face_grp, vol_grp):
 
 
 def _apply_face_mass_operator(dcoll: DiscretizationCollection, dd, vec):
-    if isinstance(vec, np.ndarray):
-        return obj_array_vectorize(
-            lambda vi: _apply_face_mass_operator(dcoll, dd, vi), vec
+    if not isinstance(vec, DOFArray):
+        return rec_map_dof_array_container(
```
alexfikl, line 751:
```suggestion
        return map_array_container(
```

grudge/projection.py
```diff
@@ -52,17 +52,15 @@ def project(dcoll: DiscretizationCollection, src, tgt, vec):
     :arg vec: a :class:`~meshmode.dof_array.DOFArray` or a
         :class:`~arraycontext.ArrayContainer`.
     """
+    if not isinstance(vec, DOFArray):
+        return rec_map_dof_array_container(
```
alexfikl, line 56:
```suggestion
        return map_array_container(
```

alexfikl (commented):
Gave this a very quick look and left a few comments here and there, hope you don't mind!

The `TracePair` stuff seems a bit more complicated, so I mostly skipped it :grin:

grudge/trace_pair.py
alexfikl: Would it be clearer make something like [unflatten_like](https://github.com/inducer/meshmode/blob/04bcfba6a94091c90d488b628a516d0316a1e5f3/meshmode/dof_array.py#L557-L612) in meshmode for this? That basically works by passing in a prototype array container (i.e. just keep a reference to the original container you flattened) and reconstructs the flat version from that, instead of keeping track of all the offsets.

It could be worth extending `flatten` (or just `flatten_to_numpy`?) with a flag to force it to flatten the complete array container, something like
```
result = []
rec_map_reduce_array_container(
	reduce_func=result.append, 
	map_func=lambda x: x, 
	flatten_to_numpy(ary))
np.concatenate(result)
```

Just some thoughts, not quite sure what the best way to handle this would be..

examples/wave/wave-op-mpi.py
```diff
@@ -51,42 +58,58 @@
...
 
+@with_container_arithmetic(bcast_obj_array=True, rel_comparison=True)
+@dataclass_array_container
+@dataclass(frozen=True)
+class WaveState:
+    u: DOFArray
+    v: np.ndarray  # [object array]
+
+    def __post_init__(self):
+        assert isinstance(self.v, np.ndarray) and self.v.dtype.char == "O"
```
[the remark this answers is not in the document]
thomasgibson: Done! https://github.com/inducer/grudge/commit/b14e2e3cc4c81658dd4a44623fabe39d631f4361

grudge/op.py
thomasgibson: Good catch, I've replaced all these recursive calls with their non-recursive counterpart: https://github.com/inducer/grudge/commit/18b8c70579801101fb01410b930a4d276a0dafcd

grudge/trace_pair.py
thomasgibson: @alexfikl Having something like `unflatten_like` would be very helpful. Unfortunately, I am not quite as familiar with the array traversal infrastructure in place as well as yourself or @inducer --- so I am not sure if something like this already exists within the current functionality.

Really at the end of the day, I have two problems here:

1. What is a good numpy representation of an arbitrary array container?
2. Define an invertible mapping from arbitrary array container to numpy representation.

My solution to 1. is to simply flatten the entire data structure. 2. is trickier because there "inverse" transformation is not so straightforward (at least to me). It sounds like the inverse mapping I want is very similar to what `unflatten_like` is doing. However, I don't recall it working right when I attempted to use it. I'll need to revisit it...

thomasgibson: Hence my reason for asking for help! :smile:

alexfikl: Yeah, the current `flatten`/`unflatten` functions don't do what you need here. They traverse the array container tree and just work on the `DOFArray`s, leaving the rest of the tree as is.

thomasgibson: Tried cleaning up in https://github.com/inducer/grudge/pull/154/commits/e37462caddf99d48c9476c14f976da9b670afec4

grudge/trace_pair.py
```diff
@@ -272,61 +274,122 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+                flat_data.append(np.concatenate(numpy_values))
+            else:
+                numpy_values = actx.to_numpy(value)
+                flat_data.append(numpy_values)
+            containerized_data_template[key] = numpy_values
+        return np.concatenate(flat_data), containerized_data_template
+    else:
+        # No data template for arys with only a single state
+        # e.g. a single DOFArray
+        return flatten_to_numpy(actx, ary), None
```
thomasgibson, line 296: @inducer @alexfikl I think this function is fairly straightforward --- it's very similar to just regular `flatten_to_numpy` with the addition that it flattens numpy object arrays of numpy arrays. This special casing is ugly, but it seems to be rampant in `meshmode` anyways, so it's a known issue.

The reason this function _also_ returns something called a `containerized_data_template` is that this is used to help unflatten and recontainerize.

grudge/trace_pair.py
```diff
@@ -272,61 +274,122 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+                    actx, discr, numpy_ary[offset:offset+_ndofs]
+                )
+                offset += _ndofs
+            remote_container_data[key] = remote_data
+        return deserialize_container(
+            template,
+            remote_container_data.items()
+        )
+    else:
+        return unflatten_from_numpy(actx, discr, numpy_ary)
```
thomasgibson, line 335: @inducer @alexfikl This is my ugly attempt and writing an inverse transformation of `flatten_to_numpy_from_container`. I think this function probably needs the most scrutiny, as there may be better ways of doing this that I haven't quite realized yet. Or this may reveal something that's missing in `meshmode.dof_array` / `arraycontext.traversal`?

thomasgibson: Note that `ary_container_data` in this function is what was previously returned as `containerized_data_template` in `flatten_to_numpy_from_container`. I should probably use consistent naming.

thomasgibson (commented):
Some comments to help explain my thoughts.

grudge/trace_pair.py
thomasgibson: Fixed a subtle bug in https://github.com/inducer/grudge/pull/154/commits/44968709360df8ed37edea973499ce257d602461

thomasgibson:
The Mirgecom CI failure seems to be due to using the `map_array_container` in the operator application routines (not the distributed memory bits). Will investigate this...

alexfikl:
> The Mirgecom CI failure seems to be due to using the `map_array_container` in the operator application routines (not the distributed memory bits). Will investigate this...
> 
> Edit: I don't know if it's something fundamentally wrong with `map_array_container`, but for some reason, the `map_array_container` is attempting to serialize floats!?

Looks like [map_array_container](https://github.com/inducer/arraycontext/blob/0f9133215e3fe6955a6da78d99fdde540b24cc3f/arraycontext/container/traversal.py#L181-L187) is not catching the `NotImplementedError` for floats correctly and trying to deserialize them anyway. I'd say this is a bug in `arraycontext`.

thomasgibson:
@alexfikl I think it was a very subtle type-checking issue! I managed to fix the mrigecom issue with this commit: https://github.com/inducer/grudge/pull/154/commits/42adaf10fe05c18141c8043b346b09f55d8554ab

grudge/trace_pair.py
thomasgibson: This commit https://github.com/inducer/grudge/pull/154/commits/9127937adb3480204192813fa5de3f80be99109a cleans things up

grudge/trace_pair.py
```diff
@@ -272,61 +277,130 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+        # of *serialize_container(template)*.
+        for key, _value in serialize_container(template):
+            # NOTE: The values of *template* determine how the values of
+            # the re-containerized data should look. For example, if we need
+            # to create an object array as one of the components.
+            # To construct the object array from a slice of flattened data,
+            # we need to inspect *_value* from *template* to determine
+            # the right offset parameter for accessing *numpy_ary*. The actual
+            # data in *_value* is not needed.
+            if isinstance(_value, np.ndarray) and _value.dtype == "O":
```
inducer, line 327: Why are object arrays handled specially here? Can't they go through the same logic as all other containers? (If not, add a comment to explain.)

grudge/trace_pair.py
```diff
@@ -272,61 +277,130 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+    numpy array suitable for communication via mpi4py.
+    """
+    if is_array_container(ary):
+        flat_data = []
+        # NOTE: The ordering of data in the flattened array will be determined
+        # by the ordering of key/value pairs returned by *serialize_container*.
+        # This is important to remember because the function
+        # *_unflatten_from_numpy_to_container* relies on this same ordering
+        # to unflatten the concatenated array.
+        for _, value in serialize_container(to_numpy(ary, actx)):
```
inducer, line 292: In order to be valid, this needs an addition to the array container serialization spec (make a PR), namely that the iterables come out in the same order every time, for types of the same cardinality. I'm not opposed to that assumption, but it's needed in order for this to be correct, and it's not currently there.

grudge/trace_pair.py
```diff
@@ -272,61 +277,130 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+            # NOTE: *to_numpy* can return numpy object arrays whose components
+            # are numpy arrays. As a result, we need to flatten these as well.
+            if isinstance(value, np.ndarray) and value.dtype == "O":
```
inducer, line 295: Why are object arrays handled specially here? Can't they go through the same logic as all other containers? (If not, add a comment to explain.)

grudge/trace_pair.py
```diff
@@ -272,61 +277,130 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+                flat_data.append(np.concatenate(value))
+            else:
+                flat_data.append(value)
+        # All data is concatenated into a single numpy array
+        flattened_np_ary = np.concatenate(flat_data)
```
inducer, line 300: For a deeply nested structure, this does a lot of recursive concatenating, amounting to a lot of copying. Instead, build a list of flattened arrays and concatenate them all at once.

inducer (commented):
Thanks for working on this! A few wrinkles (hopefully easy to answer/fix) in the communication bit, but otherwise this looks nice!

inducer:

grudge/trace_pair.py
thomasgibson: @inducer Because `serialize_container` does not serialize a nested object array inside an array container.

alexfikl: From a super quick look, this doesn't seem to be recursive. Will it handle other arbitrarily nested array containers?

grudge/trace_pair.py
thomasgibson: @inducer Because serialize_container does not serialize a nested object array inside an array container and the flattened array has absolutely no clue what array container it originally came from. It's basically what I wrote in the `# NOTE: ...`

grudge/trace_pair.py
thomasgibson: @inducer I quickly scanned through `arraycontext` to get a better understanding of how `serialize_container` is implemented to try to see if it is straightforward to extend. It's not... (to me)

Instead, I opted to just simply use an OrderedDict as a data template and perform reconstructions based on that: https://github.com/inducer/grudge/pull/154/commits/9efbe2b3b3f957fe8ae303a94186ffd6f1fa0034

Is it a good solution? Well, it seems to work just fine. I don't think it's any worse than what's already floating around. But perhaps you have a different opinion on this?

grudge/trace_pair.py
thomasgibson: @alexfikl No, but the special casing of object arrays sorta touches on this problem at bit.

alexfikl: Yeah, `serialize_container` just gives the children of one node in the container. You would need to recurse into each `value` and flatten it bottom up to avoid any special casing like that.

grudge/trace_pair.py
```diff
@@ -272,61 +281,144 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+        # If *ary* is not a container-type, then *data_template* is left empty
+        # since there is only one 'component'
+        return flatten_to_numpy(actx, ary), None
+
+
+def _unflatten_from_numpy_to_container(
+        actx: ArrayContext,
+        container_template: ArrayOrContainerT,
+        data_template: Optional[Dict[Any, Any]],
+        numpy_ary, discr) -> ArrayOrContainerT:
```
alexfikl, line 323: I'm confused why you need both a `container_template` and a `data_template` here? Wouldn't the `container_template` already prescribe how the data is laid out?

grudge/trace_pair.py
inducer: I had missed this initially (good catch @alexfikl!), but I agree: this should support nested containers in their full generality, i.e. it needs to be recursive.

grudge/trace_pair.py
inducer: I'm not sure I understand your comment here. What I was asking for was a PR into arraycontext to document that `serialize_container` produces the keys in a reliable order. If that's not the case, you can't assume that you got the keys for the template in the same order as the keys for the original data. (I.e. no code change, just codify a property that all current implementations already satisfy.)

thomasgibson: Sorry. I completely misunderstood what you were originally asking.

grudge/trace_pair.py
thomasgibson: This PR now requires https://github.com/inducer/arraycontext/pull/91

grudge/trace_pair.py
thomasgibson: Solved thanks to this :+1: https://github.com/inducer/arraycontext/pull/91

grudge/reductions.py
```diff
@@ -111,16 +113,19 @@ def norm(dcoll: DiscretizationCollection, vec, p, dd=None) -> float:
...
+            return np.fabs(vec)
+
         if p == 2:
             return sum(
-                norm(dcoll, vec[idx], p, dd=dd)**2
-                for idx in np.ndindex(vec.shape)
-            )**0.5
+                norm(dcoll, comp, p, dd=dd)**2
+                for _, comp in serialize_container(vec)
+            )**(1/2)
```
majosm, line 124: Would it make sense to use `actx.np.sqrt` here? (I changed the one in `_norm` in #170, but that PR doesn't touch this part of the code, so I didn't see this one until now.)

grudge/reductions.py
```diff
@@ -152,12 +158,15 @@ def nodal_sum_loc(dcoll: DiscretizationCollection, dd, vec) -> float:
...
     :returns: a scalar denoting the rank-local nodal sum.
     """
-    if isinstance(vec, np.ndarray):
-        return sum(nodal_sum_loc(dcoll, dd, vec[idx])
-                   for idx in np.ndindex(vec.shape))
+    if not isinstance(vec, DOFArray):
+        return sum(
+            nodal_sum_loc(dcoll, dd, comp)
+            for _, comp in serialize_container(vec)
+        )
```
majosm, line 169: I wonder if these reduction functions (and `norm`) should have an optional flag indicating whether to sum across components or not? I could imagine wanting to take a `norm` of a `ConservedVars` and get back a `ConservedVars` containing the norm of each component.

grudge/reductions.py
alexfikl: I would vote to use `sqrt` where possible, but don't feel particularly strongly about it.

majosm: Can you use `get_container_context_recursively`?

thomasgibson: Fair point :smile:

thomasgibson: > Can you use `get_container_context_recursively`?

@majosm After experimenting with this, I found a niche corner case that's mildly infuriating :laughing: here: https://github.com/inducer/grudge/blob/main/test/test_grudge.py#L1045

So it appears the original implementation (pre new world grudge) tried to be fairly general with its input. Now we're trying to assert that every `vec` being passed to `norm` has an array context. Why this isn't consistent with the other nodal reductions is likely due to an oversight on my part. (though to be honest I don't really know why we want to support the case presented [here](https://github.com/inducer/grudge/blob/main/test/test_grudge.py#L1045))

This patch makes it work consistently:
```python
if p == 2:
    from arraycontext import get_container_context_recursively

    actx = get_container_context_recursively(vec)
    if actx is not None:
        return actx.np.sqrt(
            sum(
                norm(dcoll, comp, p, dd=dd)**2
                for _, comp in serialize_container(vec)
            )
        )
    else:
        return sum(
            norm(dcoll, comp, p, dd=dd)**2
            for _, comp in serialize_container(vec)
        )**(1/2)
```

However, this special-casing is kinda ugly. I am still not solely convinced to drop `**(1/2)` when it just pollutes the code even further from my perspective (_and_ it already works as intended, no?). I am still open to being convinced.

majosm: Fair enough. I'm wondering if we should just remove the recursion on `norm`, all this square-rooting-then-squaring seems clunky to me anyway. I might take a stab at it in a separate PR.

grudge/op.py
```diff
@@ -552,10 +565,10 @@ def _apply_mass_operator(
     )
 
 
-def mass(dcoll: DiscretizationCollection, *args):
+def mass(dcoll: DiscretizationCollection, *args) -> ArrayOrContainerT:
     r"""Return the action of the DG mass matrix on a vector (or vectors)
     of :class:`~meshmode.dof_array.DOFArray`\ s, *vec*. In the case of
-    *vec* being an object array of :class:`~meshmode.dof_array.DOFArray`\ s,
+    *vec* being an :class:`~arraycontext.container.ArrayContainer`,
     the mass operator is applied in the Kronecker sense (component-wise).
```
inducer, line 572: I'm actually not sure I buy that applying componentwise is accurately described as a Kronecker product.

grudge/op.py
```diff
@@ -187,12 +190,13 @@ def local_grad(
         \nabla|_E f = \left(
             \partial_x|_E f, \partial_y|_E f, \partial_z|_E f \right)
 
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
```
inducer, line 194:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```
(here and elsewhere)

grudge/trace_pair.py
```diff
@@ -275,58 +280,49 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
-        self.bdry_discr = dcoll.discr_from_dd(self.remote_btag)
+    def __init__(self,
+                 dcoll: DiscretizationCollection,
+                 array_container: ArrayOrContainerT,
+                 remote_rank, tag=None):
+        actx = get_container_context_recursively(array_container)
+        btag = BTAG_PARTITION(remote_rank)
 
-        from grudge.op import project
+        local_array_ctr = project(dcoll, "vol", btag, array_container)
```
inducer, line 290:
```suggestion
        local_bdry_data = project(dcoll, "vol", btag, array_container)
```
Or some such. (also rename the attribute)

grudge/trace_pair.py
```diff
@@ -275,58 +280,49 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+        numpy_data = to_numpy(flatten(local_array_ctr, actx), actx)
```
inducer, line 291:
```suggestion
        local_bdry_data_np = to_numpy(flatten(local_array_ctr, actx), actx)
```

grudge/trace_pair.py
```diff
@@ -275,58 +280,49 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
-        local_data = self.array_context.to_numpy(flatten(self.local_dof_array))
-        comm = self.dcoll.mpi_communicator
+        self.tag = self.base_tag
+        if tag is not None:
+            self.tag += tag
 
-        self.send_req = comm.Isend(local_data, remote_rank, tag=self.tag)
-        self.remote_data_host = np.empty_like(local_data)
-        self.recv_req = comm.Irecv(self.remote_data_host, remote_rank, self.tag)
+        self.send_req = comm.Isend(numpy_data, remote_rank, tag=self.tag)
```
inducer, line 304: This might be a good moment to look up mpi4py semantics (and add a comment on our usage here). As it's written, `numpy_data` will go out of scope shortly after the `Isend` is posted, and it's not obvious whether that's long enough to allow the send to complete.

My quick skim of the [mpi4py docs](https://mpi4py.readthedocs.io/en/stable/overview.html#nonblocking-communications) doesn't settle this unambiguously, unfortunately.

At any rate, I think the safe approach might be to keep the buffer alive until the send completes (while adding a comment summarizing what we know/assume about the lifetime semantics).

grudge/trace_pair.py
```diff
@@ -275,58 +280,49 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
         self.recv_req.Wait()
 
+        # Re-containerize the remote data
         actx = self.array_context
-        remote_dof_array = unflatten(
-            self.array_context, self.bdry_discr,
-            actx.from_numpy(self.remote_data_host)
-        )
-
+        exchanged_data = from_numpy(self.remote_data_host_numpy, actx)
```
inducer, line 314:
```suggestion
        remote_bdry_data_flat = from_numpy(self.remote_data_host_numpy, actx)
```

grudge/trace_pair.py
```diff
@@ -275,58 +280,49 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+        remote_array_ctr = unflatten(self.local_array_ctr,
```
inducer, line 315:
```suggestion
        remote_bdry_data = unflatten(self.local_array_ctr,
```

grudge/trace_pair.py
```diff
@@ -344,38 +340,17 @@ def cross_rank_trace_pairs(
...
-            TracePair(
-                dd=dof_desc.as_dofdesc(
-                    dof_desc.DTAG_BOUNDARY(BTAG_PARTITION(remote_rank))),
-                interior=make_obj_array([
-                    result[remote_rank, i].int for i in range(n)]).reshape(oshape),
-                exterior=make_obj_array([
-                    result[remote_rank, i].ext for i in range(n)]).reshape(oshape)
-            ) for remote_rank in connected_ranks(dcoll)
-        ]
+    if isinstance(ary, Number):
```
inducer, line 347: We should probably document that we assume that if a number is passed, the same number is passed on every rank.

grudge/trace_pair.py
```diff
@@ -344,38 +340,17 @@ def cross_rank_trace_pairs(
...
+        return [TracePair(BTAG_PARTITION(remote_rank), interior=ary, exterior=ary)
+                for remote_rank in connected_ranks(dcoll)]
     else:
-        return _cross_rank_trace_pairs_scalar_field(dcoll, ary, tag=tag)
+        return [_RankBoundaryCommunication(dcoll, ary,
+                                           remote_rank, tag=tag).finish()
+                for remote_rank in connected_ranks(dcoll)]
```
inducer, line 353: Writing the code like this defeats the purpose of having `_RankBoundaryCommunication` be an object! :) Post and completion might as well be in the same function. Instead, create all the `_RankBoundaryCommunication` instances in a first pass (posting the sends and receives) and *then* wait for completion in a second pass.

inducer (commented):
Thanks for working on this, and sorry about the epic delay in getting this reviewed! I spotted one thing that I think is significant in the MPI rework (but that should be a quick fix still). The rest is superficial stuff. After these are resolved, this is ready to land from my perspective.

grudge/trace_pair.py
thomasgibson: I snuck in a clarifying comment in https://github.com/inducer/grudge/pull/154/commits/414c0b374a9fb327ad54303012af77a9bcdc16e8 :)

grudge/op.py
thomasgibson: I've gone ahead and just simplified it by referring to "component-wise" application. I think it's much more clear anyways :)

In case you're wondering where I got the idea of Kronecker products, it's [here](https://www.brown.edu/research/projects/scientific-computing/sites/brown.edu.research.projects.scientific-computing/files/uploads/Review%20of%20entropy%20stable%20discontinuous%20Galerkin%20methods.pdf), page 13

grudge/trace_pair.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/154/commits/7b0729a5e17eacae466a46e285513cb5fa62e446

grudge/trace_pair.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/154/commits/7b0729a5e17eacae466a46e285513cb5fa62e446

grudge/trace_pair.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/154/commits/7b0729a5e17eacae466a46e285513cb5fa62e446

grudge/trace_pair.py
thomasgibson: Done: https://github.com/inducer/grudge/pull/154/commits/7b0729a5e17eacae466a46e285513cb5fa62e446

grudge/trace_pair.py
thomasgibson: I've gone ahead provided some comments/notes (to the best of my understanding): https://github.com/inducer/grudge/pull/154/commits/55950b121a97fcca77e73c356dfa62bba501b3ca

thomasgibson:
@inducer Ready for another look!

thomasgibson:
@inducer when this lands, do not delete this branch; I will take care of it since I need to reset the base branch in https://github.com/inducer/grudge/pull/172

grudge/op.py
```diff
@@ -405,10 +416,14 @@ def weak_local_d_dx(dcoll: DiscretizationCollection, *args):
...
+        Defaults to the base volume discretization if not provided.
     :arg xyz_axis: an integer indicating the axis along which the derivative
-        is taken
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
-    :returns: a :class:`~meshmode.dof_array.DOFArray`\ s.
+        is taken.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
```
inducer, line 426: Are you sure that this function supports array containers?

grudge/op.py
```diff
@@ -224,8 +229,10 @@ def local_d_dx(dcoll: DiscretizationCollection, xyz_axis, vec):
 
     :arg xyz_axis: an integer indicating the axis along which the derivative
         is taken.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
-    :returns: a :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
```
inducer, line 235: Are you sure this function supports array containers?

grudge/op.py
```diff
@@ -455,11 +470,13 @@ def weak_local_div(dcoll: DiscretizationCollection, *args):
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
         Defaults to the base volume discretization if not provided.
-    :arg vec: a object array of
-        a :class:`~meshmode.dof_array.DOFArray`\ s,
+    :arg vecs: an object array of
+        :class:`~meshmode.dof_array.DOFArray`\s or
+        :class:`~arraycontext.container.ArrayContainer` objects,
```
inducer, line 475: I don't think `_div_helper` deals with anything other than object arrays.

grudge/op.py
```diff
@@ -263,19 +270,21 @@ def _div_helper(dcoll: DiscretizationCollection, diff_func, vecs):
...
-    :arg vec: an object array of
-        a :class:`~meshmode.dof_array.DOFArray`\ s,
+    :arg vecs: an object array of
+        :class:`~meshmode.dof_array.DOFArray`\ s or
+        :class:`~arraycontext.container.ArrayContainer` objects,
         where the last axis of the array must have length
         matching the volume dimension.
-    :returns: a :class:`~meshmode.dof_array.DOFArray`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
```
inducer, line 287: I don't think `_div_helper` deals with anything other than object arrays.

grudge/projection.py
```diff
@@ -32,37 +32,41 @@
...
     quadrature discretization.
 
     :arg src: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
     :arg tgt: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or a
-        :class:`~arraycontext.ArrayContainer`.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
```
inducer, line 59:
```suggestion
    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
        :class:`~arraycontext.container.ArrayContainer` of them.
    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
        :class:`~arraycontext.container.ArrayContainer` like *vec*.
```

grudge/reductions.py
```diff
@@ -78,10 +83,8 @@ def norm(dcoll: DiscretizationCollection, vec, p, dd=None) -> "DeviceScalar":
     r"""Return the vector p-norm of a function represented
     by its vector of degrees of freedom *vec*.
 
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an object array of
-        a :class:`~meshmode.dof_array.DOFArray`\ s,
-        where the last axis of the array must have length
-        matching the volume dimension.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
```
inducer, line 87:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```

grudge/reductions.py
```diff
@@ -115,7 +118,8 @@ def nodal_sum(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value
         convertible to one.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     :returns: a scalar denoting the nodal sum.
```
inducer, line 123:
```suggestion
    :returns: a device scalar denoting the nodal sum.
```

grudge/reductions.py
```diff
@@ -152,7 +159,8 @@ def nodal_min(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value
         convertible to one.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     :returns: a scalar denoting the nodal minimum.
```
inducer, line 164:
```suggestion
    :returns: a device scalar denoting the nodal minimum.
```

grudge/reductions.py
```diff
@@ -192,7 +203,8 @@ def nodal_max(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value
         convertible to one.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     :returns: a scalar denoting the nodal maximum.
```
inducer, line 208:
```suggestion
    :returns: a device scalar denoting the nodal maximum.
```

grudge/reductions.py
```diff
@@ -192,7 +203,8 @@ def nodal_max(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value
         convertible to one.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
```
inducer, line 207:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```

grudge/reductions.py
```diff
@@ -173,12 +181,15 @@ def nodal_min_loc(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
...
```
inducer, line 185:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```

grudge/reductions.py
```diff
@@ -152,7 +159,8 @@ def nodal_min(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
...
```
inducer, line 163:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```

grudge/reductions.py
```diff
@@ -135,12 +139,15 @@ def nodal_sum_loc(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
...
```
inducer, line 143:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```

grudge/reductions.py
```diff
@@ -232,7 +247,8 @@ def integral(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
     :class:`~meshmode.dof_array.DOFArray` of degrees of freedom.
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
     :returns: a scalar denoting the evaluated integral.
```
inducer, line 252:
```suggestion
    :returns: a device scalar denoting the evaluated integral.
```

grudge/reductions.py
```diff
@@ -232,7 +247,8 @@ def integral(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
     :class:`~meshmode.dof_array.DOFArray` of degrees of freedom.
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
```
inducer, line 251:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```

grudge/reductions.py
```diff
@@ -317,67 +327,121 @@ def elementwise_prg():
...
 
-    May be called with ``(dcoll, vec)`` or ``(dcoll, dd, vec)``.
+    May be called with ``(vec)`` or ``(dd, vec)``.
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
         Defaults to the base volume discretization if not provided.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`
-    :returns: a :class:`~meshmode.dof_array.DOFArray` whose entries
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
```
inducer, line 340:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them
```

grudge/reductions.py
```diff
@@ -317,67 +327,121 @@ def elementwise_prg():
...
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` whose entries
```
inducer, line 342:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` like *vec* whose entries
```

grudge/reductions.py
```diff
@@ -317,67 +327,121 @@ def elementwise_prg():
...
 
     :arg dcoll: a :class:`grudge.discretization.DiscretizationCollection`.
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
         Defaults to the base volume discretization if not provided.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`
-    :returns: a :class:`~meshmode.dof_array.DOFArray` whose entries
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` whose entries
```
inducer, line 371:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` like *vec* whose entries
```

grudge/op.py
```diff
@@ -332,26 +341,28 @@ def get_ref_stiffness_transpose_mat(out_grp, in_grp):
...
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
     :arg nested: return nested object arrays instead of a single multidimensional
         array if *vec* is non-scalar
     :returns: an object array (possibly nested) of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+        :class:`~meshmode.dof_array.DOFArray`\ s or
+        :class:`~arraycontext.container.ArrayContainer`\ s.
```
inducer, line 365:
```suggestion
        :class:`~arraycontext.container.ArrayContainer`\ s like *vec*,
        with the object array from the gradient innermost.
```

grudge/op.py
```diff
@@ -572,11 +587,10 @@ def mass(dcoll: DiscretizationCollection, *args):
...
         Defaults to the base volume discretization if not provided.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
-    :returns: a :class:`~meshmode.dof_array.DOFArray` denoting the
-        application of the mass matrix, or an object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
```
inducer, line 593:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` like *vec*.
```

grudge/op.py
```diff
@@ -686,11 +697,10 @@ def inverse_mass(dcoll: DiscretizationCollection, vec):
...
 
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
-    :returns: a :class:`~meshmode.dof_array.DOFArray` denoting the
-        application of the inverse mass matrix, or an object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
```
inducer, line 703:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` like *vec*.
```

grudge/op.py
```diff
@@ -862,11 +872,10 @@ def face_mass(dcoll: DiscretizationCollection, *args):
...
         Defaults to the base ``"all_faces"`` discretization if not provided.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
-    :returns: a :class:`~meshmode.dof_array.DOFArray` denoting the
-        application of the face mass matrix, or an object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` of them.
```
inducer, line 878:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` like *vec*.
```

grudge/reductions.py
```diff
@@ -317,67 +327,121 @@ def elementwise_prg():
         )
 
 
-def elementwise_sum(dcoll: DiscretizationCollection, *args) -> DOFArray:
+def elementwise_sum(
```
inducer, line 330: I think there's an opportunity to document the elementwise reductions more precisely: namely, mention that they return `DOFArrays` with the same number of degrees of freedom as the input, all set to the same value. Now that I type that out, that's kind of dumb, and kind of unnecessary when proper broadcasting is available. But that's for another time. -> #184 

The doc concern stands though, even in view of that direction.

grudge/reductions.py
```diff
@@ -317,67 +327,121 @@ def elementwise_prg():
...
 
     :arg dcoll: a :class:`grudge.discretization.DiscretizationCollection`.
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
         Defaults to the base volume discretization if not provided.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`
-    :returns: a :class:`~meshmode.dof_array.DOFArray` whose entries
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` whose entries
```
inducer, line 400:
```suggestion
    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
        :class:`~arraycontext.container.ArrayContainer` of them.
    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
        :class:`~arraycontext.container.ArrayContainer` like *vec* whose entries
```

grudge/reductions.py
```diff
@@ -317,67 +327,121 @@ def elementwise_prg():
...
+
+    :arg dcoll: a :class:`grudge.discretization.DiscretizationCollection`.
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray`
-    :returns: a :class:`~meshmode.dof_array.DOFArray` containing the
+        Defaults to the base volume discretization if not provided.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
+    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` containing the
```
inducer, line 430:
```suggestion
    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
        :class:`~arraycontext.container.ArrayContainer` of them.
    :returns: a :class:`~meshmode.dof_array.DOFArray` or an
        :class:`~arraycontext.container.ArrayContainer` like *vec* containing the
```

grudge/trace_pair.py
```diff
@@ -175,10 +179,12 @@ def bdry_trace_pair(
 
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one,
         which describes the boundary discretization.
-    :arg interior: a :class:`~meshmode.dof_array.DOFArray` that contains data
+    :arg interior: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` that contains data
```
inducer, line 183:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them that contains data
```

grudge/trace_pair.py
```diff
@@ -175,10 +179,12 @@ def bdry_trace_pair(
...
         already on the boundary representing the interior value to be used
         for the flux.
-    :arg exterior: a :class:`~meshmode.dof_array.DOFArray` that contains data
+    :arg exterior: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` that contains data
```
inducer, line 187:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them that contains data
```

grudge/trace_pair.py
```diff
@@ -195,19 +201,20 @@ def bv_trace_pair(
...
         defined in the volume, which will be restricted to the boundary denoted
         by *dd*. The result will be used as the interior value
         for the flux.
-    :arg exterior: a :class:`~meshmode.dof_array.DOFArray` that contains data
+    :arg exterior: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer` that contains data
         that already lives on the boundary representing the exterior value to
         be used for the flux.
     :returns: a :class:`TracePair` on the boundary.
     """
```
inducer, line 214: Add a note that `interior` and `exterior` must have the same structure if they're containers.

grudge/trace_pair.py
```diff
@@ -219,12 +226,10 @@ def _interior_trace_pair(dcoll: DiscretizationCollection, vec) -> TracePair:
     *dcoll* with a discretization tag specified by *discr_tag*.
     This does not include interior faces on different MPI ranks.
 
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
```
inducer, line 230:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```

grudge/trace_pair.py
```diff
@@ -243,8 +248,8 @@ def interior_trace_pairs(dcoll: DiscretizationCollection, vec) -> list:
     defined on the interior faces of *dcoll* and any faces connected to a
     parallel boundary.
 
-    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or object array of
-        :class:`~meshmode.dof_array.DOFArray`\ s.
+    :arg vec: a :class:`~meshmode.dof_array.DOFArray` or an
+        :class:`~arraycontext.container.ArrayContainer`.
```
inducer, line 252:
```suggestion
        :class:`~arraycontext.container.ArrayContainer` of them.
```

grudge/trace_pair.py
```diff
@@ -275,58 +280,74 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+        self.send_req = comm.Isend(self.local_bdry_data_np,
+                                   remote_rank,
+                                   tag=self.tag)
+        self.remote_data_host_numpy = np.empty_like(self.local_bdry_data_np)
+        self.recv_req = comm.Irecv(self.remote_data_host_numpy,
+                                   remote_rank,
+                                   tag=self.tag)
 
     def finish(self):
+        # Wait for the nonblocking recieve request to complete before
```
inducer, line 330:
```suggestion
        # Wait for the nonblocking receive request to complete before
```

grudge/trace_pair.py
```diff
@@ -275,58 +280,74 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+        # Here, we initialize both send and recieve operations through
+        # mpi4py `Request` (MPI_Request) instances for comm.Isend (MPI_Isend)
+        # and comm.Irecv (MPI_Irecv) respectively. These initiate non-blocking
+        # point-to-point communication requests and require explicit management
+        # via the use of wait (MPI_Wait, MPI_Waitall, MPI_Waitany, MPI_Waitsome),
+        # test (MPI_Test, MPI_Testall, MPI_Testany, MPI_Testsome), and cancel
+        # (MPI_Cancel). The rank-local data `self.local_bdry_data_np` will have its
+        # associated memory buffer sent across connected ranks and must not be
+        # modified at the Python level during this process. Completion of the
+        # requests are handled in :meth:`finish`.
```
inducer, line 314:
```suggestion
        # requests is handled in :meth:`finish`.
```

grudge/trace_pair.py
```diff
@@ -275,58 +280,74 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+        #
+        # For more details on the mpi4py semantics, see:
+        # https://mpi4py.readthedocs.io/en/stable/overview.html#nonblocking-communications
+        #
+        # NOTE: mpi4py does hold onto the memory buffer for `self.local_bdry_data_np`
+        # until the send/recv requests are complete.
```
inducer, line 320:
```suggestion
        # NOTE: mpi4py currently (2021-11-03) holds a reference to the send memory buffer for
        # (i.e. `self.local_bdry_data_np`) until the send requests is complete, however it is not clear
        # that this is documented behavior. We hold on to the buffer (via the instance attribute)
        # as well, just in case.
```

grudge/trace_pair.py
```diff
@@ -275,58 +280,74 @@ def connected_ranks(dcoll: DiscretizationCollection):
...
+        self.recv_req = comm.Irecv(self.remote_data_host_numpy,
+                                   remote_rank,
+                                   tag=self.tag)
 
     def finish(self):
+        # Wait for the nonblocking recieve request to complete before
+        # accessing the data
         self.recv_req.Wait()
 
+        # Nonblocking recieve is complete, we can now access the data and apply
```
inducer, line 334:
```suggestion
        # Nonblocking receive is complete, we can now access the data and apply
```

grudge/trace_pair.py
```diff
@@ -344,38 +365,27 @@ def cross_rank_trace_pairs(
     components, respectively. Each of the TracePair components are structured
     like *ary*.
 
-    :arg ary: a single :class:`~meshmode.dof_array.DOFArray`, or an object
-        array of :class:`~meshmode.dof_array.DOFArray`\ s
-        of arbitrary shape.
+    During the case when *ary* is a number, rather than a
+    :class:`~meshmode.dof_array.DOFArray` or an
+    :class:`~arraycontext.container.ArrayContainer` of them, it is assumed
+    that the same number is being communicated on every rank.
```
inducer, line 371:
```suggestion
    If *ary* is a number, rather than a
    :class:`~meshmode.dof_array.DOFArray` or an
    :class:`~arraycontext.container.ArrayContainer` of them, it is assumed
    that the same number is being communicated on every rank.
```

inducer (commented):
Thanks, and sorry again about the long wait on another review. Also sorry that I've uncovered a few more things. I hope most of this is easy. There are a lot of comments here, but most of them are just about consistency in doing the same thing in multiple places. I hope you'll be able to dispatch them quickly using the "apply suggestions" feature.

inducer:

grudge/op.py
thomasgibson: I believe this is all cleared up in https://github.com/inducer/grudge/pull/154/commits/a87734006d247785f3c5baf43bde860c46dbd0c9. `_div_helper` is intentionally only working on object arrays, as the container is expected to have vector (obj array) entries.

grudge/op.py
thomasgibson: I believe this is all cleared up in https://github.com/inducer/grudge/pull/154/commits/a87734006d247785f3c5baf43bde860c46dbd0c9. `_div_helper` is intentionally only working on object arrays, as the container is expected to have vector (obj array) entries.

grudge/op.py
thomasgibson: It definitely does now :smile: https://github.com/inducer/grudge/pull/154/commits/eeb80eb3554ebd3d80be6e68ca71d38d92b931aa

grudge/op.py
thomasgibson: It definitely does now :smile: https://github.com/inducer/grudge/pull/154/commits/eeb80eb3554ebd3d80be6e68ca71d38d92b931aa

grudge/trace_pair.py
thomasgibson: Done in https://github.com/inducer/grudge/pull/154/commits/e7478d4abbbcd540936c416004573b2907cfa347

grudge/reductions.py
thomasgibson: I've added notes to all the `elementwise_x` functions: https://github.com/inducer/grudge/pull/154/commits/e7478d4abbbcd540936c416004573b2907cfa347

thomasgibson:
@inducer This is ready for another look

inducer:
Thanks for working on this! Some proposed changes to this in #191. This LGTM (and is ready to land in my view) subject to those (or similar) changes.

inducer:
Thanks for seeing this through! In it goes. :tada:

PR #161, opened 2021-08-30
P0 geometry

inducer:
This defines a connection from the default nodal discretization to a P^0 (pw constant) discretization, to which all constant-per-element geometry terms (geometry derivatives on simplices) are interpolated. That results in `DOFArray`s with group arrays of shape `(nelements, 1)`, which, happily, broadcast cleanly over `(nelements, nunit_dofs)` with exactly the desired semantics. The other PRs have already wired up the infrastructure bits to make this go.

Needs (all merged):
- https://github.com/inducer/modepy/pull/42
- https://github.com/inducer/arraycontext/pull/85
- https://github.com/inducer/arraycontext/pull/86
- https://github.com/illinois-ceesd/mirgecom/pull/505

### DIFF
--- a/grudge/geometry/metrics.py
+++ b/grudge/geometry/metrics.py
@@ -22,6 +22,7 @@ Geometry terms
 Normal vectors
 --------------
 
+.. autofunction:: mv_normal
 .. autofunction:: normal
 
 Curvature tensors
--- a/grudge/geometry/metrics.py
+++ b/grudge/geometry/metrics.py
@@ -379,7 +379,8 @@ def inverse_metric_derivative(
 
 def inverse_surface_metric_derivative(
         actx: ArrayContext, dcoll: DiscretizationCollection,
-        rst_axis, xyz_axis, dd=None):
+        rst_axis, xyz_axis, dd=None,
+        *, _use_geoderiv_connection=False):
     r"""Computes the inverse surface metric derivative of the physical
     coordinate enumerated by *xyz_axis* with respect to the
     reference axis *rst_axis*. These geometric terms are used in the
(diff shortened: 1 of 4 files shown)


inducer:
I've rebased this to include and interoperate with #165 (now merged).

inducer:
Whoops. Meant to update the description before requesting review, ultimately forgot to do that. Did that just now. Sorry if the context wasn't clear.

grudge/geometry/metrics.py
```diff
@@ -391,6 +393,12 @@ def inverse_surface_metric_derivative(
     :arg xyz_axis: an integer denoting the physical coordinate axis.
     :arg dd: a :class:`~grudge.dof_desc.DOFDesc`, or a value convertible to one.
         Defaults to the base volume discretization.
+    :arg _use_geoderiv_connection: If *True*, some of the returned group arrays
```
thomasgibson, line 396: Perhaps we can give this a better name? I know mechanically it uses the new geoderivative connection, but how about calling something along the lines of `_affine_geometry` or `_use_affine_storage`?

thomasgibson (approved):
I think this looks great. Just a very minor comment, but otherwise good to go!

grudge/geometry/metrics.py
inducer: I think I'd prefer to keep the name.

First, it's internal-facing (cf. the leading underscore), so I don't mind if the name is technical. It's part of the docstring not because I'd like it to be used outside of where it is currently, but rather to describe what it does. I'll add a note to that effect.

Second, I'd like this flag to prescribe "do whatever the geoderiv thing does", instead of suggesting that "it has its own agenda."

What do you think?

inducer: https://github.com/inducer/grudge/compare/b4fd08a4de60181c247c1ce2cb121ed7d299f26b..4c2b185e739099a2abab91d52bb6b8148a91524b shows these changes. I've also made the flag defined specifically in terms of the geoderiv connection instead of blathering about P^0 for affine. To compensate, I've added the geoderiv connection to the docs (also with a warning sign about being internal) to explain what it's for.

thomasgibson: @inducer This is a perfectly fine alternative for me!

PR #165, opened 2021-09-05
More data reuse in differentiation

### LINKED ISSUE #144: `_compute_local_gradient` largely duplicates the code of `_apply_stiffness_transpose_operator`
As a result, (versions of) #143 and #142 exist here, too. I think it should be straightforward to make them the same routine with an additional parameter or two.

This is something I should have caught during review of #74.

cc @thomasgibson @lukeolson

inducer:
Still to do:

- [x] Needs https://github.com/inducer/arraycontext/pull/87
- [x] Allow premultipling metric derivatives with jacobian
- [x] Pull xyz_axis into loopy kernel in grad
- [x] Pull xyz_axis into loopy kernel in weak grad
- [x] Unify weak and strong derivatives

cc @kaushikcfd 

Closes #144.

### DIFF
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -269,38 +269,6 @@ def reference_stiffness_transpose_matrix(
                                            in_element_group)
 
 
-def _apply_stiffness_transpose_operator(
-        dcoll: DiscretizationCollection, dd_out, dd_in, vec, xyz_axis):
-    from grudge.geometry import \
-        inverse_surface_metric_derivative_mat, area_element
-
-    in_discr = dcoll.discr_from_dd(dd_in)
-    out_discr = dcoll.discr_from_dd(dd_out)
-
-    actx = vec.array_context
-    area_elements = area_element(actx, dcoll, dd=dd_in)
-    inverse_jac_mat = inverse_surface_metric_derivative_mat(actx, dcoll, dd=dd_in)
-
-    return DOFArray(
-        actx,
-        data=tuple(
-            actx.einsum("dij,ej,ej,dej->ei",
-                        reference_stiffness_transpose_matrix(
-                            actx,
-                            out_element_group=out_grp,
-                            in_element_group=in_grp
-                        ),
-                        ae_i,
-                        vec_i,
-                        ijm_i[xyz_axis],
-                        arg_names=("ref_stiffT_mat", "jac", "vec", "inv_jac_t"),
-                        tagged=(FirstAxisIsElementsTag(),))
-
-            for out_grp, in_grp, vec_i, ae_i, ijm_i in zip(
-                out_discr.groups, in_discr.groups, vec, area_elements,
-                inverse_jac_mat)))
-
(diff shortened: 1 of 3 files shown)


inducer:
@kaushikcfd This is done and headed for `main` once CI passes. I'll merge it into #161 once it's on main. It may be of interest because it avoids loading the area elements in the differentiation entirely: they get premultiplied into the geometric factors. The same optimization applies for affine, but it likely won't make as much of a difference.

cc @thomasgibson: This cleans up the derivative operators a bit. It avoids the repeated `stack`, unifies back-end implementations of weak and strong, and facilitates premultiplication of the area elements into the inverse metric derivatives.

inducer:
@thomasgibson Please do take a look and leave a review, even if it's post-merge. (It's probably easiest to read this commit-by-commit.)

inducer:
I'll actually leave this off auto-merge for the moment, until someone tells me that the `metric_in_matvec` flag from the last commit isn't insane.

inducer:
Upon sleeping on it, I think it's not too crazy. Still interested in what you think, but I'll likely merge this if it passes.

grudge/op.py
```diff
@@ -92,11 +92,68 @@
...
+                    arg_names=("inv_jac_t", "ref_stiffT_mat", "vec"),
+                    tagged=(FirstAxisIsElementsTag(),))
+        for out_grp, in_grp, vec_i, ijm_i in zip(
+            out_discr.groups, in_discr.groups, vec,
+            inv_jac_mat)]
+
+    return make_obj_array([
+            DOFArray(
+                actx, data=tuple([pgg_i[xyz_axis] for pgg_i in per_group_grads]))
+            for xyz_axis in range(out_discr.ambient_dim)])
```
thomasgibson, line 142: Maybe emphasize here in this function (as a comment or dosctring, nothing too complicated) that this routine is more or less supporting two types of differentiation (weak and strong --- depending on what gets passed in as `get_diff_mat` and `metric_in_matvec`.

thomasgibson (approved):
I'm overall fine with this. Just a minor comment.

thomasgibson:
> Upon sleeping on it, I think it's not too crazy. Still interested in what you think, but I'll likely merge this if it passes.

I don't think it's too crazy at all. The only difference is a scaling factor in the geometric terms and the elemental matrix. It's the same type of local mat vec otherwise, so I'm totally cool with this.

inducer:
Hi all, I totally fumbled the social aspect of this. My attempts to figure out whether I should just merge this or wait on a review may have ended up creating a sense of urgency that just didn't exist. I apologize if any of you felt pressured in any way, that was not the intention. Sorry!

grudge/op.py
inducer: Agreed, see 3987a0a.

inducer:
Thanks for taking a look @thomasgibson!

PR #170, opened 2021-09-24
Return device scalars from `nodal_sum`/`nodal_min`/`nodal_max`

### LINKED ISSUE #168: Reductions return a device scalar without MPI and a host scalar with MPI
We should fix that inconsistency. To avoid forcing an evaluation if lazy is in use, I think the sanest thing to do is to standardize on device scalars.

cc @kaushikcfd @majosm @thomasgibson

majosm:
Closes #168.


.github/workflows/ci.yml
```diff
@@ -93,7 +93,11 @@ jobs:
                 . ./ci-support-v0
 
                 if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
-                    git clone "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
+                    if [[ "$GITHUB_HEAD_REF" = "nodal-reduction-device-scalar" ]]; then
+                        git clone "https://github.com/majosm/$DOWNSTREAM_PROJECT.git" -b "nodal-reduction-device-scalar"
+                    else
+                        git clone "https://github.com/illinois-ceesd/$DOWNSTREAM_PROJECT.git"
+                    fi
```
thomasgibson, line 100: We'll need to drop this before merging

grudge/reductions.py
```diff
@@ -80,19 +80,21 @@ def _norm(dcoll: DiscretizationCollection, vec, p, dd):
...
         from grudge.op import _apply_mass_operator
-        return abs(
-            nodal_sum(
-                dcoll,
-                dd,
-                vec.conj() * _apply_mass_operator(dcoll, dd, dd, vec))
-            )**(1/2)
+        return vec.array_context.np.sqrt(
+            # Quantities being summed are real up to rounding error, so abs() can
+            # go on the outside
```
alexfikl, line 85: This isn't just for real numbers though, since there's a `conj` in there. So, it would seem better to either
* remove the `conj` and assert that `vec.entry_dtype` is real
* add back the `abs` to handle complex numbers too

What do you think?

.github/workflows/ci.yml
majosm: I _think_ this needs to stay for now, otherwise the mirgecom CI job will fail and I won't be able to merge. It can be removed again once illinois-ceesd/mirgecom#518 is also merged. It won't affect anything in main since the branch name isn't `nodal-reduction-device-scalar`.

grudge/reductions.py
```diff
@@ -80,19 +80,21 @@ def _norm(dcoll: DiscretizationCollection, vec, p, dd):
...
+                    dd,
+                    vec.conj() * _apply_mass_operator(dcoll, dd, dd, vec))))
     elif p == np.inf:
         return nodal_max(dcoll, dd, abs(vec))
     else:
         raise NotImplementedError("Unsupported value of p")
 
 
-def norm(dcoll: DiscretizationCollection, vec, p, dd=None) -> float:
+def norm(dcoll: DiscretizationCollection, vec, p, dd=None):
```
alexfikl, line 97:
```suggestion
def norm(dcoll: DiscretizationCollection, vec, p, dd=None) -> Any:
```
? Not that `grudge` is type checked, but `mypy` [only checks a function if there's a return type](https://mypy.readthedocs.io/en/stable/getting_started.html#more-function-signatures), so maybe for the future.

grudge/reductions.py
```diff
@@ -159,12 +162,23 @@ def nodal_sum_loc(dcoll: DiscretizationCollection, dd, vec) -> float:
...
+    if hasattr(actx, "_force_device_scalars"):
+        force_device_scalars = actx._force_device_scalars
+    else:
+        force_device_scalars = True
+
+    def device_sum(x):
+        result = actx.np.sum(x)
+        if not force_device_scalars:
+            result = actx.from_numpy(np.array(result))
+        return result
```
alexfikl, line 176: Is this extra work required? In the `not force_device_scalars` case (which is just a deprecated `pyopencl` thing), the reductions always return host scalars, but this forces the grudge reductions to return device scalars, which seems inconsistent.

Does anything break if this just does the `actx.np.sum` and uses whatever that returns? [`to_numpy`](https://github.com/inducer/arraycontext/blob/c5ed317ac595d4ee1e222930d6e0eb0b3315f366/arraycontext/impl/pyopencl/__init__.py#L164-L165) should let these pass through, hopefully.

grudge/dt_utils.py
```diff
@@ -179,12 +179,14 @@ def h_max_from_volume(
     if dim is None:
         dim = dcoll.dim
 
-    ones = dcoll.discr_from_dd(dd).zeros(dcoll._setup_actx) + 1.0
-    return nodal_max(
+    actx = dcoll._setup_actx
+
+    ones = dcoll.discr_from_dd(dd).zeros(actx) + 1.0
+    return actx.to_numpy(nodal_max(
```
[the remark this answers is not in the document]
thomasgibson: In otherwords, are there situations where we always want to force to host scalar?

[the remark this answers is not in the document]
thomasgibson: > In general, I would say that if the function got some array context (here from `dcoll`), it makes sense for it to return some array type specific to that context and, if the user wants a value in another array context (including `numpy`), they should explicitly ask for it.

I think this makes a lot of sense as a general rule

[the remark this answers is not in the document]
majosm: > Why is `hmax` (and `hmin` below) forced into a host scalar here?

Yeah, I can't say I gave much thought to what these _should_ return, I just tried to fix them up to behave as before. Making them return device scalars doesn't seem to break anything in a way that `actx.to_numpy` can't fix, though -- so I added the change here.

grudge/reductions.py
majosm: Note that I didn't actually change the `abs` here, I just added a comment to attempt to explain why it (apparently) works even though the `abs` looks like it's in the wrong place (outside the sum vs. inside).

majosm: If I understand correctly, the result of `vec.conj() * _apply_mass_operator(dcoll, dd, dd, vec)` would be complex-typed but nonnegative real-valued (up to floating point error), so doing the `abs()` on the inside or on the outside should give the same result (also up to floating point error). Is this right @thomasgibson?

thomasgibson: @majosm You are correct, though if we do decide to change our basis functions (for example using Berstein polynomials), the mass operation is no long SPD (I think) and hence you're not guaranteed to have positiveness. In which case, the correct approach is to put `abs` _inside_. So I think if we want to have something more bullet-proof, let's put `abs` inside.

Does that sound okay to you guys?

grudge/reductions.py
majosm: I took another look at this, and I think I can safely remove the `force_device_scalars` handling from `nodal_sum_loc`, but not the others. In `nodal_min_loc`/`nodal_max_loc`, I need device scalar results for each `grp_array` so they can subsequently be passed into `actx.np.minimum`/`actx.np.maximum`. I tried to tweak the return values a little in 214e865 to make them more compatible with `force_device_scalars`. Can you tell me what you think?

grudge/dt_utils.py
```diff
@@ -158,7 +159,7 @@ def dt_non_geometric_factors(
 
 @memoize_on_first_arg
 def h_max_from_volume(
-        dcoll: DiscretizationCollection, dim=None, dd=None) -> float:
+        dcoll: DiscretizationCollection, dim=None, dd=None) -> Any:
```
thomasgibson, line 162: Actually I feel like type annotating with `Any` is not very helpful. In fact, I feel it just sorta raises more questions. I'd rather just remove the return annotation (at least for `h_max` and `h_min` here).

grudge/reductions.py
```diff
@@ -75,24 +76,23 @@
 
 # {{{ Nodal reductions
 
-def _norm(dcoll: DiscretizationCollection, vec, p, dd):
+def _norm(dcoll: DiscretizationCollection, vec, p, dd) -> Any:
```
thomasgibson, line 79: Same here, I'd prefer to remove the type annotation here.

grudge/reductions.py
```diff
@@ -178,11 +179,13 @@ def nodal_min(dcoll: DiscretizationCollection, dd, vec) -> float:
...
     from mpi4py import MPI
+    actx = vec.array_context
 
-    return comm.allreduce(nodal_min_loc(dcoll, dd, vec), op=MPI.MIN)
+    return actx.from_numpy(
+        comm.allreduce(actx.to_numpy(nodal_min_loc(dcoll, dd, vec)), op=MPI.MIN))
 
 
-def nodal_min_loc(dcoll: DiscretizationCollection, dd, vec) -> float:
+def nodal_min_loc(dcoll: DiscretizationCollection, dd, vec) -> Any:
```
thomasgibson, line 188: In the case here, I think we should just document the return behavior in the docstring rather than annotate with `Any`. For this function (and similar), I think it'll be helpful to mention that the return type of these routines will depend on `force_device_scalars`

alexfikl: Just to add to this, the reason I suggested adding `Any` is that `mypy` won't type check this function at all if it doesn't have a return type, but it's not an issue at the moment, I guess.

Completely agree it's useless from a documentation point of view.

alexfikl: Also, maybe mentioning that `force_device_scalars` thing is not the best, since it's just a `PyOpenCLArrayContext` thing that's supposed to be deprecated already?

grudge/reductions.py
alexfikl: I'm definitely ok with that, was just wondering what didn't quite work. When we introduced that `force_device_scalars` thing, the hope was that `to_numpy` would just pass it through so that user code won't have to check.

In case you tried this already, what breaks if you just write, e.g.
```
    return reduce(
        lambda acc, grp_ary: actx.np.minimum(acc, actx.np.min(grp_ary)),
        vec, np.inf)
```
?

majosm: Without the `force_device_scalars` handling I was seeing errors like [these](https://github.com/inducer/grudge/runs/3861787012?check_suite_focus=true) (the `AttributeError: 'numpy.float64' object has no attribute 'queue'`).

PR #176, opened 2021-10-19
Tweak var velocity advection demo to show aliasing instability

inducer:
Run with
```
pycl var-velocity.py --visualize --flux=central
```
to see the instability. Run with
```
pycl var-velocity.py --visualize --flux=central --use-quad
```
to see how overintegration adresses the issue. In this example, it seems that upwinding has enough dissipation to overwhelm the instability and stabilize the method.

cc @MTCam

### DIFF
--- a/examples/advection/var-velocity.py
+++ b/examples/advection/var-velocity.py
@@ -99,7 +99,8 @@ class Plotter:
 # }}}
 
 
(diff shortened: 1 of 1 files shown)


thomasgibson (approved):
Nice addition to the advection example. Thanks.

PR #182, opened 2021-11-01
Avoid call_loopy in `elementwise_reduction` by taking a combination of axis-reduction and broadcast_to

kaushikcfd:


inducer:
Thanks!

PR #185, opened 2021-11-12
add lazy option to wave-op-mpi

matthiasdiener:


examples/wave/wave-op-mpi.py
```diff
@@ -141,15 +141,20 @@ def bump(actx, dcoll, t=0):
...
+    actx_outer = PyOpenCLArrayContext(
         queue,
         allocator=cl_tools.MemoryPool(cl_tools.ImmediateAllocator(queue)),
         force_device_scalars=True,
     )
 
+    if lazy:
+        actx_rhs = PytatoPyOpenCLArrayContext(queue)
+    else:
+        actx_rhs = actx_outer
```
thomasgibson, line 156: Rather than have two different `actx` floating around, why not just strictly use the pytato context throughout the script when in lazy-mode? I would just kill `actx_outer` entirely. It should work just fine, as I do that in my ESDG examples (which strictly use lazy): https://github.com/inducer/grudge/blob/ESDG-development/examples/euler/acoustic_pulse.py#L236

inducer: I agree with that. While it's nice to be able to use two array contexts, it's kind of clunky.

inducer:
The CI appears to be failing for now. Request review when it's ready for a look or @-mention if it needs attention.

matthiasdiener:
This is ready for review @inducer @thomasgibson

examples/wave/wave-op-mpi.py
```diff
@@ -192,14 +196,16 @@ def main(ctx_factory, dim=2, order=3, visualize=False):
...
+
     if comm.rank == 0:
         logger.info("dt = %g", dt)
 
     t = 0
     t_final = 3
     istep = 0
     while t < t_final:
-        fields = rk4_step(fields, t, dt, rhs)
+        fields = rk4_step(fields, t, dt, compiled_rhs)
```
thomasgibson, line 208: I think we might need to still perform the `thaw/freeze` pattern as illustrated in meshmode's simple-dg example. Maybe @inducer has more to say on this?
```suggestion
        fields = thaw(freeze(fields, actx), actx)
        fields = rk4_step(fields, t, dt, compiled_rhs)
```

Actually, I think my suggestion might not be the optimal way. The right way is probably illustrated here: https://github.com/inducer/meshmode/blob/main/examples/simple-dg.py#L494-L496 --- meaning we just wrap the rhs evaluations in a thaw/freeze pattern.

thomasgibson (commented):
Just one last comment, then this looks good to go. It's just concerning the thaw/freeze pattern I've been conditioned to do :sweat_smile:

examples/wave/wave-op-mpi.py
matthiasdiener: Hmm, but isn't the thaw/freeze pattern due to using two different actx (like simple-dg does? 

I'm also confused why the ESDG example just thaws/freezes on the same actx:
https://github.com/inducer/grudge/blob/49bcb38ba6de002eeec8fb383998877339e7a753/examples/euler/acoustic_pulse.py#L209-L210

thomasgibson: I was under the impression that this was necessary to keep pytato's compiler from doing some unnecessary/repeated code generation between timesteps (which is an incredibly frustrating issue itself). I recall this being an issue for `mirgecom` as well for @MTCam's temperature solver. Perhaps I am the one misunderstanding then. Would really appreciate @kaushikcfd or @inducer to enlighten us on this.

If I _don't_ do the thaw/freeze in between calls to the timestepper, I experienced repeated code-generation that kills the simulation runtime. That's the summary of my issue. And I would expect that this example likely also suffers from that. I haven't checked out this branch to run it (yet) but are you not experiencing this problem with the example as is?

thomasgibson: I was under the impression that this was necessary to keep pytato's compiler from doing some unnecessary/repeated code generation between timesteps (which is an incredibly frustrating issue itself). I recall this being an issue for `mirgecom` as well for the temperature solver. Perhaps I am the one misunderstanding then

If I _don't_ do the thaw/freeze in between calls to the timestepper, I experienced repeated code-generation that kills the simulation runtime. That's the summary of my issue. And I would expect that this example likely also suffers from that. I haven't checked out this branch to run it (yet) but are you not experiencing this problem with the example as is?

matthiasdiener: I see, you're right! Should be fixed in 3e710a8

.github/workflows/ci.yml
```diff
@@ -64,8 +64,6 @@ jobs:
                 build_py_project_in_conda_env
                 run_examples
 
-                python wave/wave-op-mpi.py --lazy
```
thomasgibson: @matthiasdiener Out of curiosity, what went wrong when attempting to run this? Also does https://github.com/inducer/grudge/commit/3e710a8c1ff85ad1ae1eeadabff6bcc8ce5887ca resolve this now?

thomasgibson (commented):
Last question on testing (sorry :sweat_smile:) Otherwise I'm ready to merge this.

.github/workflows/ci.yml
matthiasdiener: I didn't want to introduce a hard dependency on pytato, also, the latest released version of pytato doesn't seem to be able to run this code.

kaushikcfd: I don't think we need to introduce a hard dependency on `pytato` for this, we can add the git repo to `requirements.txt`.

matthiasdiener: Good idea, I added it back to the tests

thomasgibson (commented):
LGTM. Thanks @matthiasdiener!

examples/wave/wave-op-mpi.py
```diff
@@ -192,14 +196,19 @@ def main(ctx_factory, dim=2, order=3, visualize=False):
...
     if comm.rank == 0:
         logger.info("dt = %g", dt)
 
     t = 0
     t_final = 3
     istep = 0
     while t < t_final:
-        fields = rk4_step(fields, t, dt, rhs)
+        if lazy:
+            fields = thaw(freeze(fields, actx), actx)
```
inducer, line 209: This is odd, and it shouldn't be here. What happens if you remove it?

thomasgibson: This is directly related to this discussion: https://github.com/inducer/grudge/pull/185#discussion_r757025159. Unresolved so we can continue the discussion there.

examples/wave/wave-op-mpi.py
kaushikcfd: For the PytatoPyOpenCLArrayContext, if we don't do this the fields starts becoming expression as below:
```
# at istep = 1
fields = fields(0) + dt/6*(k1(0) + 2*k2(0) + 2*k3(0)+k4(0))
# at istep = 2
fields = fields(0) + dt/6*(k1(0) + 2*k2(0) + 2*k3(0)+k4(0)) + dt/6*(k1(dt) + 2*k2(dt) + 2*k3(dt) + k4(dt))
```
I.e. the expression keeps growing in size. This is a consequence of us not thunking the values onto the expression objects. I agree this is a sub-optimal user-interface (Very related: https://github.com/inducer/arraycontext/issues/102).

PR #186, opened 2021-11-12
Do not deprecate rank-local interior trace pairs

thomasgibson:
This simple PR removes the deprecation warning for `interior_trace_pair`, since there are legitimate reasons (cc. @MTCam and the Navier-Stokes operator) you'd want to handle the rank-local interior trace pairs independently from the cross-rank pairs.

Also we may want to try some latency hiding in lazy mode soon, so we probably want to treat all rank-local surface + volume terms as a unit when attempting to overlap communication with computation.

### DIFF
--- a/grudge/trace_pair.py
+++ b/grudge/trace_pair.py
@@ -16,8 +16,8 @@ Boundary trace functions
 Interior and cross-rank trace functions
 ---------------------------------------
 
-.. autofunction:: interior_trace_pair
 .. autofunction:: interior_trace_pairs
+.. autofunction:: local_interior_trace_pair
 .. autofunction:: cross_rank_trace_pairs
 """
 
@@ -215,7 +215,7 @@ def bv_trace_pair(
 
 # {{{ Interior trace pairs
 
-def interior_trace_pair(dcoll: DiscretizationCollection, vec) -> TracePair:
+def local_interior_trace_pair(dcoll: DiscretizationCollection, vec) -> TracePair:
     r"""Return a :class:`TracePair` for the interior faces of
     *dcoll* with a discretization tag specified by *discr_tag*.
     This does not include interior faces on different MPI ranks.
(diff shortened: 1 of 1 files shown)


inducer:
Out of curiosity: What's the use case for maintaining the local-only trace pairs? Could you add a comment (to the code) to explain why we need to keep both facilities?

MTCam:
> Out of curiosity: What's the use case for maintaining the local-only trace pairs? Could you add a comment (to the code) to explain why we need to keep both facilities?

tbh, I don't care about the local-only trace pairs; only the partition-boundary-only trace pairs.  I sometimes process those differently (e.g. computing temperature on the partition boundary instead of communicating it).  Eventually, I end up glomming the PB and local-only trace pairs together as "interior" trace pairs.

Flipping this around, what's the impetus for eliminating the ability to pull down local-only pairs?

inducer:
> Flipping this around, what's the impetus for eliminating the ability to pull down local-only pairs?

No big :axe: to grind here. Just having ported a few DG operators to MPI and wondered why we don't default to a function that unifies cross-rank and local.

FWIW, your explanation with temperatures makes good sense to me. @thomasgibson Could you add a comment to that effect?

thomasgibson:
@inducer At the moment, there's not much immediate gain other than what @MTCam already mentioned about the temperature seeds. However, thinking forward, we may want to try overlap communication with computation in a `rhs` function (in say `mirgecom`). To do this with DG, the recipe is pretty standard. You post the send/recv requests followed by computing rank-local contributions (this includes volume _and_ rank-interior trace pairs). Then you wait/synchronize and compute the cross-rank contributions. This could potentially be done at the python `rhs` level, which then distributed/parallel-enable pytato can work with.

I can add a short comment on its potential uses, sure.

inducer:
>  we may want to try overlap communication with computation in a `rhs` function (in say `mirgecom`).

Excellent point, though as far as lazy eval is concerned, the order of statements in the code has no bearing on the actual execution order, other than to maintain proper data flow. As such, lazy eval can enable that without any special accommodation in the code.

MTCam:
> > Flipping this around, what's the impetus for eliminating the ability to pull down local-only pairs?
> 
> No big 🪓 to grind here. Just having ported a few DG operators to MPI and wondered why we don't default to a function that unifies cross-rank and local.
> 
I love this notion.  It makes great sense, but would like to have the capability to do pb-specific processing. But now that @thomasgibson mentions it....

> FWIW, your explanation with temperatures makes good sense to me. @thomasgibson Could you add a comment to that effect?

Turns out, since boundary temperature seeding, we now communicate T, so it is in the regular graph.   We could probably safely transition to the built-in pb being handled automatically.  So about the capability for pb-specific processing; it is more of a want than a need at present.

Before boundary temperature seeding with the boundary-projected temperature field, I used to just communicate CV because my local version of Temperature(CV_pb)  agreed with my neighbors, but since seeding now the local Temperature(CV_pb, remote_seed) requires the remote's seed temperature to ensure agreement across the PB.

thomasgibson:
@inducer I've pushed a short comment. Hopefully, it is more or less what you were thinking.

grudge/trace_pair.py
```diff
@@ -214,11 +215,19 @@ def bv_trace_pair(
 
 # {{{ Interior trace pairs
 
-def _interior_trace_pair(dcoll: DiscretizationCollection, vec) -> TracePair:
+def interior_trace_pair(dcoll: DiscretizationCollection, vec) -> TracePair:
```
inducer, line 218: I'm not loving the naming now. `interior_trace_pair` is local-only, vs `interior_trace_pairs` is also remote? It's not obvious to me. Maybe make the un-deprecated `interior_trace_pair` available as `local_interior_trace_pair`?

inducer (commented):
Thanks! One naming quibble, then we should be good to go here.

inducer:

grudge/trace_pair.py
thomasgibson: Good call, the name was bothering me as well :smile: I'll update soon!

thomasgibson: Done (included doc updates): https://github.com/inducer/grudge/pull/186/commits/97078b6a8fb4bd2121869c17342186ae47c8f5aa

inducer:
LGTM, thanks!

PR #187, opened 2021-11-21
Make models.wave compatible with all array contexts

kaushikcfd:


thomasgibson (approved):
Thanks!

grudge/shortcuts.py
```diff
@@ -23,7 +23,7 @@
 """
 
 
-def set_up_rk4(field_var_name, dt, fields, rhs, t_start=0):
+def set_up_rk4(field_var_name, dt, fields, rhs, t_start=0.0):
```
inducer, line 26: @kaushikcfd Out of curiosity: How come this mattered?

kaushikcfd: `t` is one of the arguments to the `rhs`. Passing an `int64` type at first and then `float64` leads to compiling 2 expression graphs. This was adding up to the setup times in my experiments and hence got rid of it.

inducer: Ah. Fair enough, thanks for explaining!

PR #191, opened 2021-12-01
More shared code in array container support

inducer:
This came about as part of my review of #154. Rather than clumsily describe what I figured should happen, I thought I'd just make these changes. @thomasgibson If you agree with these changes, then #154 (which this targets) with this merged in is good to go from my end.


thomasgibson (approved):
Thanks, @inducer! I think this looks great

PR #192, opened 2021-12-07
Patch for `_div_helper`: pass dof descriptor when mapping containers

thomasgibson:
I accidentally missed this when reviewing #191 --- sorry!


inducer:
Thx!

PR #194, opened 2021-12-07
Refactor interpolation steps inside metric computations

### LINKED ISSUE #174: `DiscretizationCollection._base_to_geoderiv_connection` is broken when using overintegration in lazy-mode
The relatively recent change for storing affine geometric terms as constants for the pytato array context is broken when using overintegration. The code currently assumes that the `from_discr` is interpolatory, which is just not true when you have terms defined on a quadrature grid.

I have attempted to circumvent this issue in https://github.com/inducer/grudge/pull/172 but I would really like to settle this once and for all.  And by this, I mean actually taking advantage of affineness even when overintegrating. Especially since having working overintegration with lazy evaluation is a high priority right now.

@inducer --- I don't know if I'm being dense here, but I don't see this as a simple 2-line change in the [method](https://github.com/inducer/grudge/blob/2af3528391bf4cb07f5aacf43f7c6afcc18f693a/grudge/discretization.py#L327). The problem is getting `meshmode` to do the right thing by transferring a `DOFArray` defined on a quadrature grid to the appropriate shape for the `geo_deriv_discr`. Is there something we can do in `grudge` to spoon-feed `meshmode` something it can work with?

thomasgibson:
This resolves #174 and is a replacement for the same fix in #172

### DIFF
--- a/grudge/geometry/metrics.py
+++ b/grudge/geometry/metrics.py
@@ -114,7 +114,11 @@ def forward_metric_nth_derivative(
     if dd is None:
         dd = DD_VOLUME
 
-    inner_dd = dd.with_discr_tag(DISCR_TAG_BASE)
+    if dd.discretization_tag is not DISCR_TAG_BASE:
+        raise ValueError(
+            "The DOF descriptor associated with computing metric derivatives "
+            "must have `DISCR_TAG_BASE` as a discretization tag."
+        )
 
     if isinstance(ref_axes, int):
         ref_axes = ((ref_axes, 1),)
@@ -128,21 +132,15 @@ def forward_metric_nth_derivative(
     if len(set(ref_axes)) != len(ref_axes):
         raise ValueError("ref_axes must not contain an axis more than once")
 
-    from pytools import flatten
-    flat_ref_axes = flatten([rst_axis] * n for rst_axis, n in ref_axes)
-
     from meshmode.discretization import num_reference_derivative
 
-    vec = num_reference_derivative(
-        dcoll.discr_from_dd(inner_dd),
-        flat_ref_axes,
-        thaw(dcoll.discr_from_dd(inner_dd).nodes(), actx)[xyz_axis]
+    return num_reference_derivative(
+        dcoll.discr_from_dd(dd),
+        [rst_axis
+         for rst_axis, n in ref_axes
+         for _ in range(n)],
+        thaw(dcoll.discr_from_dd(dd).nodes(), actx)[xyz_axis]
     )
-
-    if dd.uses_quadrature():
-        vec = dcoll.connection_from_dds(inner_dd, dd)(vec)
-
-    return vec
 
 
 def forward_metric_derivative_vector(
@@ -408,20 +406,24 @@ def inverse_surface_metric_derivative(
     if dd is None:
         dd = DD_VOLUME
     dd = dof_desc.as_dofdesc(dd)
+    dd_base = dd.with_discr_tag(DISCR_TAG_BASE)
 
     if ambient_dim == dim:
         result = inverse_metric_derivative(
-            actx, dcoll, rst_axis, xyz_axis, dd=dd
+            actx, dcoll, rst_axis, xyz_axis, dd=dd_base
         )
     else:
-        inv_form1 = inverse_first_fundamental_form(actx, dcoll, dd=dd)
+        inv_form1 = inverse_first_fundamental_form(actx, dcoll, dd=dd_base)
         result = sum(
             inv_form1[rst_axis, d]*forward_metric_nth_derivative(
-                actx, dcoll, xyz_axis, d, dd=dd
+                actx, dcoll, xyz_axis, d, dd=dd_base
             ) for d in range(dim))
 
     if _use_geoderiv_connection:
-        result = dcoll._base_to_geoderiv_connection(dd)(result)
+        result = dcoll._base_to_geoderiv_connection(dd_base)(result)
+    else:
+        if dd.uses_quadrature():
+            result = dcoll.connection_from_dds(dd_base, dd)(result)
 
     return result
 
@@ -450,6 +452,9 @@ def inverse_surface_metric_derivative_mat(
         inverse metric derivatives in per-group arrays of shape
         ``(xyz_dimension, rst_dimension, nelements, ndof)``.
     """
+    if dd is None:
+        dd = DD_VOLUME
+    dd = dof_desc.as_dofdesc(dd)
 
     @memoize_in(dcoll, (inverse_surface_metric_derivative_mat, dd,
         times_area_element, _use_geoderiv_connection))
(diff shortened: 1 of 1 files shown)


grudge/geometry/metrics.py
```diff
@@ -114,7 +114,11 @@ def forward_metric_nth_derivative(
     if dd is None:
         dd = DD_VOLUME
 
-    inner_dd = dd.with_discr_tag(DISCR_TAG_BASE)
+    if dd.discretization_tag is not DISCR_TAG_BASE:
+        raise ValueError(
+            "The DOF descriptor associated with computing metric derivatives "
+            "must have `DISCR_TAG_BASE` as a discretization tag."
```
inducer, line 120: Why is this the case? I could see a rationale for asking for metric derivaties on non-base discretizations. This seems like a feature removal whose rationale I don't fully understand. Maybe add a comment/improve the error message to explain?

grudge/geometry/metrics.py
```diff
@@ -408,20 +406,24 @@ def inverse_surface_metric_derivative(
...
+            actx, dcoll, rst_axis, xyz_axis, dd=dd_base
         )
     else:
-        inv_form1 = inverse_first_fundamental_form(actx, dcoll, dd=dd)
+        inv_form1 = inverse_first_fundamental_form(actx, dcoll, dd=dd_base)
         result = sum(
             inv_form1[rst_axis, d]*forward_metric_nth_derivative(
-                actx, dcoll, xyz_axis, d, dd=dd
+                actx, dcoll, xyz_axis, d, dd=dd_base
             ) for d in range(dim))
```
inducer, line 420: I think these are nonlinear expressions and thus I think it's better to evaluate them on the quadrature grid. If you agree, maybe add a comment to that effect?

grudge/geometry/metrics.py
```diff
@@ -567,14 +572,19 @@ def area_element(
...
     if dd is None:
         dd = DD_VOLUME
+    dd = dof_desc.as_dofdesc(dd)
 
     @memoize_in(dcoll, (area_element, dd, _use_geoderiv_connection))
     def _area_elements():
+        dd_base = dd.with_discr_tag(DISCR_TAG_BASE)
         result = actx.np.sqrt(
-            pseudoscalar(actx, dcoll, dd=dd).norm_squared())
+            pseudoscalar(actx, dcoll, dd=dd_base).norm_squared())
```
inducer, line 581: This is nonlinear, too... I think that should stay on the non-base `dd`.

inducer (commented):
Thanks for working on this! I'm not sure I fully understand the reasoning here, but maybe I'm missing something. Could you explain?

thomasgibson:
> I'm not sure I fully understand the reasoning here, but maybe I'm missing something. Could you explain?

@inducer This was already discussed a bit in https://github.com/inducer/grudge/issues/174 and is the main reason for this PR in the first place. I'll do my best to summarize here as well.

The main issue is that the existing connection machinery in meshmode is quite awful at handling moving data on the quadrature grid to the "geometry discretization". The reason is that it's not an interpolatory group and the connections rely on having immediate access to a group basis.

Here is why this is actually kinda critical that we get this taken care of **ASAP**: Without these changes, overintegration and lazy evaluation in `mirgecom` is fundamentally broken ever since the introduction of this PR: https://github.com/inducer/grudge/pull/161.

The purpose here is to move the interpolation to the quadrature grid later down the stack to avoid the problems caused by this function: https://github.com/inducer/grudge/blob/ESDG-flux-refactor/grudge/discretization.py#L327-L370

I have thought about what might need to happen to fix this without modifying the metrics, but it seems like it would require a substantial amount of time to fix. If you disagree, then please let's discuss.

grudge/geometry/metrics.py
thomasgibson: From https://github.com/inducer/grudge/pull/194#issuecomment-1004189393: the reason for this change was that I moved the interpolation step out of this function and into the relevant routines in `area_element` and similar (you already commented on those parts of the code below).

grudge/geometry/metrics.py
thomasgibson: > I think it's better to evaluate them on the quadrature grid. If you agree, maybe add a comment to that effect?

I mean I would agree in principle if it was even possible to do this in `meshmode`. The metric derivatives (even before this PR) are always computed on the nodal grid since `num_reference_derivative` can only be evaluated on the interpolation grid.

In short, `meshmode` reference derivative computations are only ever computed on the nodal grid, see: https://github.com/inducer/meshmode/blob/main/meshmode/discretization/__init__.py#L614-L618. Hence why `dd_base` is always used to compute the metric derivatives.

So this comment doesn't really make much sense given our current infrastructure. All I have done in this PR is move where the interpolation happens.

grudge/geometry/metrics.py
thomasgibson: https://github.com/inducer/grudge/pull/194/commits/a7bcf7d0c41f813562d89d21df66108bb8783b03

inducer:
> awful at handling moving data on the quadrature grid to the "geometry discretization".

Could you point me to where/why that's needed? It's a bit surprising to me, since the "geometry *derivative* discretization" is either identical to base or P^0. As you point out, it's not totally clear how that should happen, as it's not guaranteed to be exact.

If we're using P^0 for geometry derivatives, there is IMO no need for metrics or derivatives to ever go to the quadrature grid, as they're all constants. So there should not be a "detour" through the quadrature grid, the "geo deriv" grid should be used (in that case, and only in that case!) where the quad grid is being used now. In the non-P^0 case, I believe the existing approach of "upsample before nonlinearities" is correct.

> Without these changes, overintegration and lazy evaluation in mirgecom is fundamentally broken

As in, it will not execute, or will silently compute garbage?

thomasgibson:
> As in, it will not execute, or will silently compute garbage?

The main issue is how we move metric terms to the geometry discretization. When affine, you interpolate to a "P0" DOFArray --- see https://github.com/inducer/grudge/blob/main/grudge/discretization.py#L316-L372. When you're not overintegrating, this is just taking metric data (on the interpolatory grid) and interpolating to P0. BUT when you're overintegrating, the metric data is not on an interpolatory grid; it's on the quadrature grid. So meshmode throws a fit because it cannot perform the interpolation.

This is what I'm trying to get around. Is the problem clear?

thomasgibson:
> > Without these changes, overintegration and lazy evaluation in mirgecom is fundamentally broken
> 
> As in, it will not execute, or will silently compute garbage?

@inducer It will not execute. It will fail every time `_base_to_geoderiv_connection` is called with a `dd` that is not interpolatory (which happens in `main` whenever you overintegrate). This is why I am not so sure it's even _possible_ to do everything on the quad grid and _then_ apply the `_base_to_geoderiv_connection` thing to the result (which seems to be what you are suggesting).

In fact, I'm now toying with the idea of removing it _entirely_ at this point, as the whole reason this was put in place was to allow for us to store a single value per element for affine geometry groups... Do we honestly need the whole `_base_to_geoderiv_connection` thing to accomplish this? Can't we just do something like:
```python
    if actx.supports_nonscalar_broadcasting:
        return DOFArray(
            actx,
            data=tuple(
                vec_i.reshape(-1, 1)
                for vec_i in vec
            )
        )
```

inducer:
Sorry for being slow to respond here, still on the road.

> This is why I am not so sure it's even possible (with the current code) to do everything on the quad grid and then apply the _base_to_geoderiv_connection thing to the result (which seems to be what you are suggesting, but maybe I'm just misunderstanding!).

Not what I'm trying to suggest, and I agree that that wouldn't work. (And I'm not sure there's a mathematically valid way either.) Instead, what I was trying to suggest was the following: For affinely-mapped element groups, no geometry information should ever leave the "geo deriv" discretization, even if data on a quadrature grid is requested. The broadcasted constant value is exact, after all.

As for an overall approach here, I have the following proposal: Make a new PR with a failing test, and I'll push a fix.

thomasgibson:
Closing in favor of https://github.com/inducer/grudge/pull/210

PR #195, opened 2021-12-07
Force a single-value per element for elementwise reductions

### LINKED ISSUE #184: Elementwise reductions shouldn't return the same value in all DOFs of the element
That's kind of a waste of space (and memory bandwidth, and, currently, compute). In the presence of proper broadcasting, a single value per element should suffice and do the same job just as well. (i.e. shape `(nelements, 1)`)

We should come up with a backward-compatible plan to get us to that point without too much pain.

cc @thomasgibson @majosm

thomasgibson:
This PR fixes #184

The logic here is that it's safe to force the elementwise reductions to have a shape of `(nelem, 1)` by definition of a reduction in this context; no need to broadcast to a full DOFArray

### DIFF
--- a/grudge/dt_utils.py
+++ b/grudge/dt_utils.py
@@ -271,16 +271,24 @@ def dt_geometric_factors(
     dd_face = DOFDesc("all_faces", dd.discretization_tag)
     face_discr = dcoll.discr_from_dd(dd_face)
 
+    # To get a single value for the total surface area of a cell, we
+    # take the sum over the averaged face areas on each face.
+    # NOTE: The face areas are the *same* at each face nodal location.
+    # This assumes there are the *same* number of face nodes on each face.
     surface_areas = abs(
         op.elementwise_integral(
             dcoll, dd_face, face_discr.zeros(actx) + 1.0
         )
     )
-
-    # Sum the areas of each face to get the total surface area
     surface_areas = DOFArray(
         actx,
         data=tuple(
+            # NOTE: Whenever the array context can't perform nonscalar
+            # broadcasting, elementwise reductions
+            # (like `elementwise_integral`) repeat the *same* scalar value of
+            # the reduction at each degree of freedom. To get a single
+            # value for the face area (per face),
+            # we simply average over the nodes, which gives the desired result.
             actx.einsum(
                 "fej->e",
                 face_ae_i.reshape(
@@ -289,12 +297,8 @@ def dt_geometric_factors(
                     -1
                 ),
                 tagged=(FirstAxisIsElementsTag(),)) / (
-                    # NOTE: Whenever the array context can't perform nonscalar
-                    # broadcasting, elementwise reductions
-                    # (like `elementwise_integral`) repeat the scalar value of
-                    # the reduction at each degree of freedom. To get a single
-                    # value for the total surface area of a cell,
-                    # we sum over all nodes and average.
+                    # NOTE: This assumes there are the *same* number of
+                    # face nodes on each face.
                     1 if actx.supports_nonscalar_broadcasting
                     else afgrp.nunit_dofs
                 )
--- a/grudge/dt_utils.py
+++ b/grudge/dt_utils.py
@@ -271,31 +271,52 @@ def dt_geometric_factors(
     dd_face = DOFDesc("all_faces", dd.discretization_tag)
     face_discr = dcoll.discr_from_dd(dd_face)
 
-    # To get a single value for the total surface area of a cell, we
-    # take the sum over the averaged face areas on each face.
-    # NOTE: The face areas are the *same* at each face nodal location.
-    # This assumes there are the *same* number of face nodes on each face.
     surface_areas = abs(
         op.elementwise_integral(
             dcoll, dd_face, face_discr.zeros(actx) + 1.0
         )
     )
-    surface_areas = DOFArray(
-        actx,
-        data=tuple(
-            actx.einsum("fej->e",
-                        face_ae_i.reshape(
-                            vgrp.mesh_el_group.nfaces,
-                            vgrp.nelements,
-                            afgrp.nunit_dofs
-                        ),
-                        tagged=(FirstAxisIsElementsTag(),)) / afgrp.nunit_dofs
 
-            for vgrp, afgrp, face_ae_i in zip(volm_discr.groups,
-                                              face_discr.groups,
-                                              surface_areas)
+    if actx.supports_nonscalar_broadcasting:
+        # Sum the areas of each face to get the total surface area
+        surface_areas = DOFArray(
+            actx,
+            data=tuple(
+                actx.einsum(
+                    "fej->e",
+                    face_ae_i.reshape(
+                        vgrp.mesh_el_group.nfaces,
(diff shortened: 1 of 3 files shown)


grudge/reductions.py
```diff
@@ -268,11 +268,17 @@ def integral(dcoll: DiscretizationCollection, dd, vec) -> "DeviceScalar":
...
-    r"""Returns a vector of DOFs with all entries on each element set
-    to the reduction operation *op_name* over all degrees of freedom.
+    r"""Returns an array container whose entries contain
+    the elementwise reductions in each cell.
 
     May be called with ``(vec)`` or ``(dd, vec)``.
 
+    Note that for array contexts which support nonscalar broadcasting
+    (e.g. :class:`meshmode.array_context.PytatoPyOpenCLArrayContext`),
+    the size of each component vector will be of shape `(nelements, 1)`.
```
kaushikcfd, line 278: Minor RST formatting:
```suggestion
    the size of each component vector will be of shape ``(nelements, 1)``.
```

kaushikcfd (approved):
LGTM, thanks!

grudge/reductions.py
thomasgibson: Thanks for catching this!

grudge/dt_utils.py
```diff
@@ -271,31 +271,52 @@ def dt_geometric_factors(
...
+                        vgrp.mesh_el_group.nfaces,
+                        vgrp.nelements,
+                        afgrp.nunit_dofs
+                    ),
+                    tagged=(FirstAxisIsElementsTag(),)) / afgrp.nunit_dofs
+
+                for vgrp, afgrp, face_ae_i in zip(volm_discr.groups,
+                                                  face_discr.groups,
+                                                  surface_areas)
+            )
```
kaushikcfd, line 318: These both branches could be integrated if we called
```python
face_ae_i.reshape(
    vgrp.mesh_el_group.nfaces,
    vgrp.nelements,
    -1
)
```

thomasgibson: Good call --- fixed: https://github.com/inducer/grudge/pull/195/commits/db98f0dff37cccef7f8d3216057487a4456171fc

kaushikcfd (approved):
Thanks!

thomasgibson:
@inducer CI is now failing for some mysterious reason ever since https://github.com/inducer/grudge/pull/200. Here's a link to the log: https://github.com/inducer/grudge/runs/4638364194?check_suite_focus=true#step:3:606. It's not clear why this (and all other PRs) are experiencing problems related to this.

inducer:
There were a few legitimate CI failures yesterday (which I fixed), but these here look spurious. I reran the checks.

grudge/dt_utils.py
```diff
@@ -271,25 +271,33 @@ def dt_geometric_factors(
     dd_face = DOFDesc("all_faces", dd.discretization_tag)
     face_discr = dcoll.discr_from_dd(dd_face)
 
-    # To get a single value for the total surface area of a cell, we
-    # take the sum over the averaged face areas on each face.
-    # NOTE: The face areas are the *same* at each face nodal location.
-    # This assumes there are the *same* number of face nodes on each face.
```
inducer, line 277: The updated code no longer describes this implicit assumption, which AFAICT is still being made.

* Why is this assumption needed in the first place?
* Could it be checked using code?

grudge/dt_utils.py
```diff
@@ -271,25 +271,33 @@ def dt_geometric_factors(
...
+                ),
+                tagged=(FirstAxisIsElementsTag(),)) / (
+                    # NOTE: Whenever the array context can't perform nonscalar
+                    # broadcasting, elementwise reductions
+                    # (like `elementwise_integral`) repeat the scalar value of
+                    # the reduction at each degree of freedom. To get a single
+                    # value for the total surface area of a cell,
+                    # we sum over all nodes and average.
+                    1 if actx.supports_nonscalar_broadcasting
+                    else afgrp.nunit_dofs
```
inducer, line 299: This nodal averaging is super sketchy to me. Why not just use honest quadrature (using `afgrp.weights` or some such) to compute areas?

inducer (commented):
Thanks! The main piece of this LGTM, however I have some questions about the area estimates in the time step estimation. If you'd rather deal with those separately (by filing an issue), I can live with that. I do think it's worth talking about this code briefly now, though.

grudge/dt_utils.py
thomasgibson: @inducer I agree this is a bit ugly, but let me explain why this was done.

You recall the issues with nonscalar broadcasting for array contexts right? Well when an array context can't support it, you get repeated values at each node, rather than the single `DOFArray` with one value per cell. This tried to address this difference.

> Why not just use honest quadrature (using afgrp.weights or some such) to compute areas?

Um, that's what is happening if you look up a few more lines. In the definition of `surface_areas`: https://github.com/inducer/grudge/blob/be1e72405c4bdb9ec7dc27462bfd77001b96c193/grudge/dt_utils.py#L274

inducer (commented):
Oops. I had a review comment pending here that I never submitted. My bad! At any rate, LGTM once that's resolved.

grudge/dt_utils.py
thomasgibson: Sounds good! Yeah I like that, I'll update accordingly

thomasgibson: Updated: https://github.com/inducer/grudge/pull/195/commits/4f04de2e48b829aca1050e298aaf5d6d7a75237e

grudge/dt_utils.py
thomasgibson: Clean up in https://github.com/inducer/grudge/pull/195/commits/4f04de2e48b829aca1050e298aaf5d6d7a75237e

grudge/dt_utils.py
```diff
@@ -271,31 +271,50 @@ def dt_geometric_factors(
...
+    if actx.supports_nonscalar_broadcasting:
+        # Compute total surface area of an element by summing over the
+        # individual face areas
+        surface_areas = DOFArray(
+            actx,
+            data=tuple(
+                actx.einsum(
+                    "fej->e",
+                    face_ae_i.reshape(
+                        vgrp.mesh_el_group.nfaces, vgrp.nelements, -1),
```
inducer, line 290:
```suggestion
                    "fe->e",
                    face_ae_i.reshape(
                        vgrp.mesh_el_group.nfaces, vgrp.nelements),
```

inducer (approved):
Thanks! I've got one possible simplification to suggest. LGTM with or without that, let me know how you feel about it.

grudge/dt_utils.py
thomasgibson: @inducer I tried this suggestion but unfortunately, I am running into this `ValueError` when I run the unit tests in `test/test_dt_utils.py`: https://github.com/inducer/pytato/blob/main/pytato/array.py#L1059 --- I think the routine is still expecting 3 indices in the einsum signature

inducer: Fair enough, thanks for trying.

PR #201, opened 2022-01-02
Fix test_surface_divergence_theorem with gmsh v4.9

alexfikl:
The test started failing with v4.9 and I couldn't find a set of "CharacteristicLengthMax" that would make it cleanly pass on v4.9 and v4.8. 

This modifies it to just use `generate_sphere` with `affine_map` to make a spheroid. The test seems to pass nicely now.

### DIFF
--- a/test/test_grudge.py
+++ b/test/test_grudge.py
@@ -33,6 +33,7 @@ pytest_generate_tests = pytest_generate_tests_for_array_contexts(
 
 from arraycontext.container.traversal import thaw
 
+from meshmode import _acf           # noqa: F401
 from meshmode.dof_array import flat_norm
 import meshmode.mesh.generation as mgen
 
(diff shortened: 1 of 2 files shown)


test/test_grudge.py
```diff
@@ -627,17 +628,17 @@ def f(x):
...
-
         # compute max element size
         from grudge.dt_utils import h_max_from_volume
 
-        h_max = h_max_from_volume(dcoll)
+        h_max = actx.to_numpy(h_max_from_volume(dcoll))
+        err_global = actx.to_numpy(abs(op_global))
+        err_local = actx.to_numpy(op.norm(dcoll, op_local, np.inf))
+        logger.info("errors: h_max %.5e global %.5e local %.5e",
+                h_max, err_global, err_local)
```
alexfikl, line 638: This is unrelated, but the `actx.to_numpy` was done after the `logger.info` and the logging didn't appreciate getting a `cl.Array` to print, so I moved it around a bit.

alexfikl:
The `mirgecom` failure is very likely unrelated. Some out-of-memory issue? I re-ran it a couple of times and it seems to persist.

inducer:
Thanks, LGTM!

PR #208, opened 2022-01-11
Use P^0 discr on affine groups even if quad grid requested

inducer:
cc @thomasgibson


thomasgibson:
This patch is incorporated in https://github.com/inducer/grudge/pull/210

PR #207, opened 2022-01-11
Overintegration + lazy evaluation is broken

thomasgibson:
This PR demonstrates the problem when using lazy evaluation _and_ overintegration. Two solutions/patches are proposed in #203 and #194 

@inducer

PR #203, opened 2022-01-05
Simplify affine-storage for geometric terms

thomasgibson:

PR #210, opened 2022-01-12
Fix metric computation for affine + overintegrated

### LINKED ISSUE #174: `DiscretizationCollection._base_to_geoderiv_connection` is broken when using overintegration in lazy-mode

thomasgibson:
Replaces #208, #207, #203, and #194 and fixes #174

### DIFF
--- a/examples/wave/wave-op-mpi.py
+++ b/examples/wave/wave-op-mpi.py
@@ -183,7 +183,8 @@ def bump(actx, dcoll, t=0):
             / source_width**2))
 
 
-def main(ctx_factory, dim=2, order=3, visualize=False, lazy=False, use_quad=False):
+def main(ctx_factory, dim=2, order=3,
+         visualize=False, lazy=False, use_quad=False, use_nonaffine_mesh=False):
     cl_ctx = ctx_factory()
     queue = cl.CommandQueue(cl_ctx)
 
@@ -205,11 +206,20 @@ def main(ctx_factory, dim=2, order=3, visualize=False, lazy=False, use_quad=Fals
     nel_1d = 16
 
     if mesh_dist.is_mananger_rank():
-        from meshmode.mesh.generation import generate_regular_rect_mesh
-        mesh = generate_regular_rect_mesh(
-                a=(-0.5,)*dim,
-                b=(0.5,)*dim,
-                nelements_per_axis=(nel_1d,)*dim)
+        if use_nonaffine_mesh:
+            from meshmode.mesh.generation import generate_warped_rect_mesh
+            # FIXME: *generate_warped_rect_mesh* in meshmode warps a
+            # rectangle domain with hard-coded vertices at (-0.5,)*dim
+            # and (0.5,)*dim. Should extend the function interface to allow
+            # for specifying the end points directly.
+            mesh = generate_warped_rect_mesh(dim=dim, order=order,
+                                             nelements_side=nel_1d)
+        else:
+            from meshmode.mesh.generation import generate_regular_rect_mesh
+            mesh = generate_regular_rect_mesh(
+                    a=(-0.5,)*dim,
+                    b=(0.5,)*dim,
+                    nelements_per_axis=(nel_1d,)*dim)
 
         logger.info("%d elements", mesh.nelements)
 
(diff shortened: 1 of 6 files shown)


examples/wave/wave-op-mpi.py
```diff
@@ -203,8 +222,21 @@ def main(ctx_factory, dim=2, order=3, visualize=False, lazy=False):
...
-    dcoll = DiscretizationCollection(actx, local_mesh, order=order,
-                    mpi_communicator=comm)
+    from meshmode.discretization.poly_element import \
+            QuadratureSimplexGroupFactory, \
+            default_simplex_group_factory
+    dcoll = DiscretizationCollection(
+        actx, local_mesh,
+        discr_tag_to_group_factory={
+            DISCR_TAG_BASE: default_simplex_group_factory(base_dim=dim, order=order),
+            DISCR_TAG_QUAD: QuadratureSimplexGroupFactory(2*order + 1),
```
inducer, line 232: Add a comment justifying the quadrature order.

examples/wave/wave-op-mpi.py
```diff
@@ -91,14 +94,29 @@ def wave_flux(dcoll, c, w_tpair):
...
     )
 
-    return op.project(dcoll, w_tpair.dd, "all_faces", c*flux_weak)
+    return op.project(dcoll, dd, dd.with_dtag("all_faces"), c*flux_weak)
 
 
-def wave_operator(dcoll, c, w):
-    u = w.u
-    v = w.v
+def wave_operator(dcoll, c, w, quad_tag=None):
```
inducer, line 100: How hard would it be to also test with a non-affine mesh?

inducer (commented):
Thanks for working on this! A few (hopefully easy) things to resolve, then this is good to go.

examples/wave/wave-op-mpi.py
thomasgibson: IMO this is out of scope and I'd rather not mess too much further with the existing examples.

thomasgibson: Nvm, I think I came up with something nonaffine for the wave op example :slightly_smiling_face: 
image

thomasgibson:
@inducer we have sporadic CI failures for downstream mirgecom testing: https://github.com/inducer/grudge/runs/4797419815?check_suite_focus=true#step:3:897 :grimacing:

examples/wave/wave-op-mpi.py
thomasgibson: https://github.com/inducer/grudge/pull/210/commits/d468071d582918d3f3c7b6962bd0168ac5560fb7

examples/wave/wave-op-mpi.py
thomasgibson: https://github.com/inducer/grudge/pull/210/commits/d468071d582918d3f3c7b6962bd0168ac5560fb7

inducer:
I think it might be a good idea to make an issue to discuss these, but I'm not sure they have much to do with grudge.

test/test_grudge.py
```diff
@@ -33,7 +33,6 @@
 
 from arraycontext.container.traversal import thaw
 
-from meshmode import _acf           # noqa: F401
```
alexfikl, line 36: For a bit of context, this was just here for debugging, e.g. if you want to run a single test with
```
python test_grudge.py 'test_something(_acf, param, param, param)'
```
instead of going through `pytest` somehow.

thomasgibson: Ah I've wondered what that did... I've gone ahead and put it back: https://github.com/inducer/grudge/pull/210/commits/286614f7c220f9a8524553526eb97094e59e005b

test/test_metrics.py
```diff
@@ -0,0 +1,108 @@
...
+                    0.05*np.cos(10*x[0])
+                    + 1.3*x[1] + np.sin(x[1]))
+            if len(x) == 3:
+                result[2] = x[2]
+            return result
+
+        from meshmode.mesh.processing import map_mesh
+        mesh = map_mesh(mesh, m)
+
+    dcoll = DiscretizationCollection(actx, mesh, order=4)
```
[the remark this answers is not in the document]
inducer: That works for me. I was worried that the WADG inverse might not be accurate enough to pass the test, but I guess that worry was misplaced.

inducer:
Tweaked the docstrings in 1212a8e. LGTM. Thanks for working on this!

Issue #211, opened 2022-01-13
Random CI failures: downstream mirgecom testing


thomasgibson:
I've noticed that grudge's CI will sporadically fail when running downstream mirgecom tests. Here is an [example](https://github.com/inducer/grudge/runs/4798352500?check_suite_focus=true#step:3:896) log file. What makes this strange is that it is not consistent and grudge will often pass CI without any issues. #210's commit history shows examples of commits where CI will bounce between pass, fail, and pass again. The commits in question did _not_ have the capacity to affect numerical computations/tests.

This one is tricky to reproduce (as in I'm unable to... 😬). The log seems to indicate a test node being taken down somehow, but I'm not quite sure what might be causing that... Logging this issue here for reference.

inducer:
Closing provisionally based on #212.

PR #212, opened 2022-01-13
Don't run mirgecom downstream CI in parallel

inducer:
Potentially address #211.

cc @thomasgibson

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,6 +115,9 @@ jobs:
                     # can't turn off MPI in mirgecom
                     export CONDA_ENVIRONMENT=conda-env.yml
                     echo "- mpi4py" >> "$CONDA_ENVIRONMENT"
+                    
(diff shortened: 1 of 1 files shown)


inducer:
Meshmode does the same and has never had an issue, FWIW:

https://github.com/inducer/meshmode/blob/19ad308ee95e3ce78b9928620e5177fb1de08cfc/.ci/install-downstream.sh#L27

inducer:
@matthiasdiener I don't fully recall. Is mirgecom expected to pass CI with pytest running parallel?

thomasgibson (approved):
LGTM! If it works for meshmode, then I think it's fine for grudge to do the same

matthiasdiener:
> @matthiasdiener I don't fully recall. Is mirgecom expected to pass CI with pytest running parallel?

I think it is expected to pass, but sometimes runs into memory issues if I remember correctly.

PR #216, opened 2022-01-28
Patch for integral computations when using overintegration

thomasgibson:


grudge/reductions.py
```diff
@@ -77,6 +77,40 @@
...
+        from grudge.geometry import area_element
+        from meshmode.transform_metadata import FirstAxisIsElementsTag
+
+        jacobians = area_element(
+            actx, dcoll, dd=dd,
+            _use_geoderiv_connection=actx.supports_nonscalar_broadcasting)
+        return DOFArray(
+            actx,
+            data=tuple(
+                actx.einsum("ei,i,ei->ei",
```
inducer, line 99:
```suggestion
                # FIXME: Using einsum only for the benefit of eager eval.
                # With proper broadcasting, this should likely just be
                # a multiplication.
                actx.einsum("ei,i,ei->ei",
```

PR #218, opened 2022-02-22
Simplify distributed-lazy array context usage

inducer:
- [x] Needs https://github.com/inducer/pytato/pull/264


matthiasdiener:
cc #183

PR #221, opened 2022-03-02
test mirgecom examples in CI

matthiasdiener:
cc: @MTCam 

Could help with catching https://github.com/illinois-ceesd/mirgecom/issues/617

cc: #222

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,6 +132,13 @@ jobs:
                 build_py_project_in_conda_env
 
                 if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
+                    # Test main branch
+                    examples/run_examples.sh ./examples
+                    
+                    # Test production branch
+                    curl -O -L https://raw.githubusercontent.com/illinois-ceesd/emirge/main/switch-mirgecom.sh
+                    chmod a+x switch-mirgecom.sh
+                    ./switch-mirgecom.sh prod
                     examples/run_examples.sh ./examples
                 fi
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,9 +133,11 @@ jobs:
 
                 if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
                     # Test main branch
+                    # Only test examples, not pytest tests, as those take too much time.
                     examples/run_examples.sh ./examples
+                else
+                    test_py_project
                 fi
 
-                test_py_project
 
 # vim: sw=4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,10 +130,11 @@ jobs:
                     sed -i "/mpi4py/ d" requirements.txt
                 fi
                 build_py_project_in_conda_env
-                test_py_project
 
                 if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
                     examples/run_examples.sh ./examples
                 fi
+
+                test_py_project
 
 # vim: sw=4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,4 +132,8 @@ jobs:
                 build_py_project_in_conda_env
                 test_py_project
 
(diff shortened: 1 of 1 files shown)


matthiasdiener:
This is ready for review @inducer @MTCam. Note that it doesn't check production, just main.

MTCam (approved):
Awesome!  This'll help! 👍

.github/workflows/ci.yml
```diff
@@ -131,12 +131,6 @@ jobs:
                 if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
                     # Test main branch
                     examples/run_examples.sh ./examples
-                    
-                    # Test production branch
-                    curl -O -L https://raw.githubusercontent.com/illinois-ceesd/emirge/main/switch-mirgecom.sh
-                    chmod a+x switch-mirgecom.sh
-                    ./switch-mirgecom.sh prod
-                    examples/run_examples.sh ./examples
```
MTCam:
```suggestion
                    
                    # Test production drivers
                    . .ci_support/production_testing_env.sh
                    . .ci_support/merge-install-production-branch.sh .
                    . .ci_support/production-drivers-install.sh .
                    . .ci_support/production-drivers-run.sh .
```

MTCam: This should run the production tests  kind of in the same sort of way as the regular examples are exercised.   The big difference here is that the "production" branch will be merged first.

matthiasdiener: Added in #224.

.github/workflows/ci.yml
```diff
@@ -127,6 +127,12 @@ jobs:
                     sed -i "/mpi4py/ d" requirements.txt
                 fi
                 build_py_project_in_conda_env
+
+                if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
+                    # Test main branch
+                    examples/run_examples.sh ./examples
+                fi
```
inducer, line 134: Maybe this should be a separate CI job? mirgecom is already the longest job in grudge, I'm reluctant to make CI turnaround even more sluggish.

MTCam: Then maybe drop the downstream `pytest` (which takes a long time)?  Running the examples (typically done within 10 minutes) will be enough and much faster.  The examples won't work if *MIRGE-Com* is broken, and they will also fail if they get anything outside the accepted solution range.

matthiasdiener: [5f5ff6b](https://github.com/inducer/grudge/pull/221/commits/5f5ff6b44eafdfb35ccd0570a60b16f3ecc4f414) implements Mike's suggestion

MTCam: > [5f5ff6b](https://github.com/inducer/grudge/pull/221/commits/5f5ff6b44eafdfb35ccd0570a60b16f3ecc4f414) implements Mike's suggestion

Cool, thanks!  Looks like this cuts the downstream testing expense in half.  Running the examples is more important than running the `pytest` tests.  Having this test in-place would have caught the example failures caused by #222 merge.

.github/workflows/ci.yml
```diff
@@ -127,6 +127,14 @@ jobs:
                     sed -i "/mpi4py/ d" requirements.txt
                 fi
                 build_py_project_in_conda_env
-                test_py_project
+
+                if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
+                    # Test main branch
+                    # Only test examples, not pytest tests, as those take too much time.
```
inducer, line 133: Why not add a separate job for examples instead of not running mirgecom pytest at all?

MTCam: Imo, running the `pytest` suite for the downstream project is of little value.   It also takes a really long time (only 30 minutes for mirgecom@main, but pushing an hour+ once NS+AV tests are added).   I think making sure the examples run and get the correct answers is a sufficient amount of testing to claim the current development does not break mirgecom.

inducer: OK, fair. You're on the receiving end of those potentially breaking changes, so if you say it's OK, I'm happy to trust you and take my shorter CI times. :)

PR #224, opened 2022-03-02
also test production drivers

matthiasdiener:


.github/workflows/ci.yml
```diff
@@ -126,13 +126,21 @@ jobs:
...
+                test_py_project
+                cd ..
 
                 if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
                     # Test main branch
                     examples/run_examples.sh ./examples
+
+                    # Test production drivers
+                    source .ci_support/production_testing_env.sh
+                    source .ci_support/merge-install-production-branch.sh .
```
MTCam, line 140: Down inside there there is a `pip install -r requirements.txt` after merging with `production`.  Will that mess stuff up, or switch the `grudge` branch?

.github/workflows/ci.yml
```diff
@@ -126,13 +126,21 @@ jobs:
...
+
                 build_py_project_in_conda_env
+                test_py_project
+                cd ..
 
                 if test "$DOWNSTREAM_PROJECT" = "mirgecom"; then
                     # Test main branch
                     examples/run_examples.sh ./examples
+
+                    # Test production drivers
```
inducer, line 138: Maybe this should be a separate CI job? mirgecom is already the longest job in grudge, I'm reluctant to make CI turnaround even more sluggish.

inducer (commented):
I'm not sure I'm sold. I'm happy to keep mirgecom main working. Production, by definition, depends on a lot of unmerged stuff (including, potentially, grudge changes), so I feel this would be a high-maintenance proposition.

inducer:
> I'm not sure I'm sold. I'm happy to keep mirgecom main working.

As an added bonus, this provides an incentive to not let production and main drift too far apart.

.github/workflows/ci.yml
MTCam: Then maybe drop the downstream `pytest` (which takes a long time)?  Running the examples (typically done within 10 minutes) will be enough and much faster.  The examples won't work if *MIRGE-Com* is broken, and they will also fail if they get anything outside the accepted solution range.

MTCam:
> > I'm not sure I'm sold. I'm happy to keep mirgecom main working.
> 
> As an added bonus, this provides an incentive to not let production and main drift too far apart.

I don't think neglecting production testing in grudge CI will affect the amount of drift between mirgecom@production and mirgecom@main.   That said, I do not disagree with your notion leave it out.

PR #226, opened 2022-03-03
Get_attr from world

MTCam:
It is OK to use COMM_WORLD here, and this continues to work even for array contexts that don't have the mpi_communicator attribute.  This fixes many failing tests in MIRGE-Com.


inducer:
Thx!

PR #230, opened 2022-03-03
get_reasonable_array_context_class improvements

matthiasdiener:
- add documentation
- add logging
- check for correct loopy branch

### DIFF
--- a/grudge/array_context.py
+++ b/grudge/array_context.py
@@ -323,7 +323,7 @@ register_pytest_array_context_factory("grudge.pytato-pyopencl",
 def get_reasonable_array_context_class(
         lazy: bool = True, distributed: bool = True
         ) -> Type[ArrayContext]:
-    """Returns the most reasonable class :class:`PyOpenCLArrayContext` currently
+    """Returns a reasonable :class:`ArrayContext` currently
     supported given the constraints of *lazy* and *distributed*."""
     if lazy:
         # lazy, non-distributed
--- a/grudge/array_context.py
+++ b/grudge/array_context.py
@@ -329,6 +329,13 @@ def get_reasonable_array_context_class(
     """Returns a reasonable :class:`PyOpenCLArrayContext` currently
     supported given the constraints of *lazy* and *distributed*."""
     if lazy:
+        if not _HAVE_SINGLE_GRID_WORK_BALANCING:
+            from warnings import warn
+            warn("No device-parallel actx available, execution will be slow."
+                 "Please make sure you have the right branches for loopy "
+                 "(https://github.com/kaushikcfd/loopy/tree/pytato-array-context-transforms) "  # noqa
+                 "and meshmode "
+                 "(https://github.com/kaushikcfd/meshmode/tree/pytato-array-context-transforms).")  # noqa
         # lazy, non-distributed
         if not distributed:
             if _HAVE_SINGLE_GRID_WORK_BALANCING:
--- a/grudge/array_context.py
+++ b/grudge/array_context.py
@@ -49,10 +49,20 @@ try:
     # https://github.com/kaushikcfd/meshmode/tree/pytato-array-context-transforms)
     from meshmode.array_context import SingleGridWorkBalancingPytatoArrayContext
 
-    # Crude check if we have the correct loopy branch
-    # (https://github.com/kaushikcfd/loopy/tree/pytato-array-context-transforms)
-    from loopy.codegen.result import get_idis_for_kernel  # noqa
-    _HAVE_SINGLE_GRID_WORK_BALANCING = True
+    try:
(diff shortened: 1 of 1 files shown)


grudge/array_context.py
```diff
@@ -317,23 +323,32 @@ class PytestPyOpenCLArrayContextFactoryWithHostScalars(
 def get_reasonable_array_context_class(
         lazy: bool = True, distributed: bool = True
         ) -> Type[ArrayContext]:
+    """Returns the most reasonable class :class:`PyOpenCLArrayContext` currently
+    supported given the constraints of *lazy* and *distributed*."""
```
inducer, line 327:
```suggestion
    """Returns a reasonable :class:`ArrayContext` currently
    supported given the constraints of *lazy* and *distributed*."""
```

grudge/array_context.py
```diff
@@ -39,10 +39,16 @@
...
+logger = logging.getLogger(__name__)
+
 try:
     # FIXME: temporary workaround while SingleGridWorkBalancingPytatoArrayContext
     # is not available in meshmode's main branch
     from meshmode.array_context import SingleGridWorkBalancingPytatoArrayContext
+
+    # Crude check if we have the correct loopy branch:
+    from loopy.codegen.result import get_idis_for_kernel  # noqa
     _HAVE_SINGLE_GRID_WORK_BALANCING = True
```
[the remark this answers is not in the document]
matthiasdiener: Added loopy branch in [bf8844a](https://github.com/inducer/grudge/pull/230/commits/bf8844ab0bdbdc6b34ebdc82cd658001ee0cf837). Could you clarify please what you mean with adding logging for this case? It already logs when _HAVE_SINGLE_GRID_WORK_BALANCING isn't true, saying `device-parallel=False` in the log string.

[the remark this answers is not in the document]
inducer: What I mean is that this should point out why single-grid-blah isn't workable, so that users have a chance to fix it.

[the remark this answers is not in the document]
matthiasdiener: Added a warning for this case in f4955ea. This situation is pretty catastrophic performance-wise, so maybe a warning is more appropriate than just logging.

grudge/array_context.py
```diff
@@ -317,23 +326,39 @@ class PytestPyOpenCLArrayContextFactoryWithHostScalars(
...
+    """Returns a reasonable :class:`PyOpenCLArrayContext` currently
+    supported given the constraints of *lazy* and *distributed*."""
     if lazy:
+        if not _HAVE_SINGLE_GRID_WORK_BALANCING:
+            from warnings import warn
+            warn("No device-parallel actx available, execution will be slow."
+                 "Please make sure you have the right branches for loopy "
+                 "(https://github.com/kaushikcfd/loopy/tree/pytato-array-context-transforms) "  # noqa
+                 "and meshmode "
+                 "(https://github.com/kaushikcfd/meshmode/tree/pytato-array-context-transforms).")  # noqa
```
inducer, line 338: That's not what I meant, and I don't think this makes sense. This will warn anytime lazy-distributed is run with meshmode `main`, which IMO is silly: that's a perfectly fine way to do business. The problem arises when someone has the right meshmode branch and a loopy branch that doesn't support it.

matthiasdiener: > That's not what I meant, and I don't think this makes sense. This will warn anytime lazy-distributed is run with meshmode `main`, which IMO is silly: that's a perfectly fine way to do business.

But wouldn't that run single-threaded on a device, ie, really slow?

> The problem arises when someone has the right meshmode branch and a loopy branch that doesn't support it.

Ok, I'll add a warning for this case.

inducer: > But wouldn't that run single-threaded on a device, ie, really slow?

If the device is a GPU, yes. For a CPUs and testing, it's not currently parallel but moderately sensible?

matthiasdiener: > > The problem arises when someone has the right meshmode branch and a loopy branch that doesn't support it.
> 
> Ok, I'll add a warning for this case.

Added a warning in 41c5e98.

> > But wouldn't that run single-threaded on a device, ie, really slow?
> 
> If the device is a GPU, yes. For a CPUs and testing, it's not currently parallel but moderately sensible?

I'm not sure. I would guess this case (ie, both loopy and meshmode on their main branches) is not too uncommon, and since "everything works lazily+distributed", it is relatively easy to overlook that this will be quite slow for most use cases, even on CPU. The warning seems to me like a good compromise, and might incentivize us to merge those branches a bit quicker ;-).

inducer: > might incentivize us to merge those branches a bit quicker ;-).

I see what you're doing there. :)

inducer:
Thx!

PR #232, opened 2022-03-03
Generalize div/grad helpers so that logic can be reused

majosm:
A step towards illinois-ceesd/mirgecom#618.

### DIFF
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -159,19 +159,19 @@ def _gradient_kernel(actx, out_discr, in_discr, get_diff_mat, inv_jac_mat, vec,
 
 # {{{ common derivative "helpers"
 
-def _div_helper(ambient_dim, component_div, scalar_type, vecs):
+def _div_helper(ambient_dim, component_div, is_scalar, vecs):
     if not isinstance(vecs, np.ndarray):
         # vecs is not an object array -> treat as array container
         return map_array_container(
-            partial(_div_helper, ambient_dim, component_div, scalar_type), vecs)
+            partial(_div_helper, ambient_dim, component_div, is_scalar), vecs)
 
     assert vecs.dtype == object
 
-    if vecs.size and not isinstance(vecs[(0,)*vecs.ndim], scalar_type):
+    if vecs.size and not is_scalar(vecs[(0,)*vecs.ndim]):
         # vecs is an object array containing further object arrays
         # -> treat as array container
         return map_array_container(
-            partial(_div_helper, ambient_dim, component_div, scalar_type), vecs)
+            partial(_div_helper, ambient_dim, component_div, is_scalar), vecs)
 
     if vecs.shape[-1] != ambient_dim:
         raise ValueError("last/innermost dimension of *vecs* argument doesn't match "
@@ -188,7 +188,7 @@ def _div_helper(ambient_dim, component_div, scalar_type, vecs):
         return result
 
 
-def _grad_helper(ambient_dim, component_grad, scalar_type, vecs, nested):
+def _grad_helper(ambient_dim, component_grad, is_scalar, vecs, nested):
     if isinstance(vecs, np.ndarray):
         # Occasionally, data structures coming from *mirgecom* will
         # contain empty object arrays as placeholders for fields.
@@ -205,16 +205,16 @@ def _grad_helper(ambient_dim, component_grad, scalar_type, vecs, nested):
         # derivatives by stacking the results.
         grad = obj_array_vectorize(
             lambda el: _grad_helper(
-                ambient_dim, component_grad, scalar_type, el, nested=nested), vecs)
+                ambient_dim, component_grad, is_scalar, el, nested=nested), vecs)
         if nested:
             return grad
         else:
             return np.stack(grad, axis=0)
 
-    if not isinstance(vecs, scalar_type):
+    if not is_scalar(vecs):
         return map_array_container(
             partial(
-                _grad_helper, ambient_dim, component_grad, scalar_type,
+                _grad_helper, ambient_dim, component_grad, is_scalar,
                 nested=nested),
             vecs)
 
@@ -283,7 +283,7 @@ def local_grad(
     return _grad_helper(
         dcoll.ambient_dim,
         partial(_strong_scalar_grad, dcoll, dd_in),
-        DOFArray,
+        lambda v: isinstance(v, DOFArray),
         vec,
         nested=nested)
 
@@ -341,7 +341,11 @@ def local_div(dcoll: DiscretizationCollection, vecs) -> ArrayOrContainerT:
             local_d_dx(dcoll, i, vec_i)
             for i, vec_i in enumerate(vec))
 
-    return _div_helper(dcoll.ambient_dim, component_div, DOFArray, vecs)
+    return _div_helper(
+        dcoll.ambient_dim,
+        component_div,
+        lambda v: isinstance(v, DOFArray),
+        vecs)
 
 # }}}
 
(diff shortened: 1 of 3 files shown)


majosm:
illinois-ceesd/mirgecom#619 is now successfully using these helpers to support array containers in symbolic `div`/`grad`.

grudge/op.py
```diff
@@ -159,84 +159,66 @@ def _gradient_kernel(actx, out_discr, in_discr, get_diff_mat, inv_jac_mat, vec,
...
-def _div_helper(dcoll, diff_func, *args):
-    if len(args) == 1:
-        vecs, = args
-        dd = dof_desc.DOFDesc("vol", dof_desc.DISCR_TAG_BASE)
-    elif len(args) == 2:
-        dd, vecs = args
-    else:
-        raise TypeError("invalid number of arguments")
-
+def _div_helper(ambient_dim, component_div, is_scalar, vecs):
```
inducer, line 162: Since I figure you're about to use these from mirgecom (i.e. outside of grudge), why not:
- move these to `grudge.tools`,
- make them public and give them better names,
- document them?

majosm: Refactored some more. Thoughts?

majosm:
@inducer Getting an unrelated(?) CI failure in docs. Any ideas?

alexfikl:
> @inducer Getting an unrelated(?) CI failure in docs. Any ideas?

That's my bad! I moved the `ElementGroupFactory` in inducer/meshmode#305 :\ I'll make a PR to fix it ASAP.

inducer (approved):
Thanks for working on this! How is d28beb6d82aaada7871bc1256e8289f7fcbd9216 for you? If you like it (and if I managed to not break CI), then this is ready to go (i.e. squash) from my end.

grudge/tools.py
```diff
@@ -235,4 +239,105 @@ def build_jacobian(actx, f, base_state, stepsize):
     return mat
 
 
+def map_subarrays(
+        f: Callable[[np.ndarray], np.ndarray],
```
majosm, line 243: `f` can also map scalars to arrays or arrays to scalars; not sure how best to indicate this with type hints here.

grudge/tools.py
```diff
@@ -235,4 +239,105 @@ def build_jacobian(actx, f, base_state, stepsize):
...
+                for idx in np.ndindex(base_shape):
+                    result[idx + out_slice] = f(ary[idx + in_slice])
+                return result
+
+
+def rec_map_subarrays(
+        f: Callable[[np.ndarray], np.ndarray],
+        in_shape: Tuple[int, ...],
+        out_shape: Tuple[int, ...],
+        ary: np.ndarray, *,
```
majosm, line 321:
```suggestion
        ary: ArrayContainer, *,
```
?

majosm, line 321:
```suggestion
        ary: ArrayContainer, *,
```
?

grudge/tools.py
```diff
@@ -235,4 +239,105 @@ def build_jacobian(actx, f, base_state, stepsize):
...
+                f"array of shape {ary.shape} is incompatible with function "
+                f"expecting input shape {in_shape}")
+        base_shape = ary.shape[:ary.ndim-len(in_shape)]
+        if len(base_shape) == 0:
+            return f(ary)
+        else:
+            in_slice = tuple(slice(0, n) for n in in_shape)
+            out_slice = tuple(slice(0, n) for n in out_shape)
+            if return_nested:
+                result = np.empty(base_shape, dtype=ary.dtype)
```
majosm, line 306:
```suggestion
                result = np.empty(base_shape, dtype=object)
```
I think this one at least needs to be an object array, since the entries going into it may be arrays.

grudge/tools.py
```diff
@@ -235,4 +239,105 @@ def build_jacobian(actx, f, base_state, stepsize):
...
+                for idx in np.ndindex(base_shape):
+                    result[idx] = f(ary[idx + in_slice])
+                return result
+            else:
+                result = np.empty(base_shape + out_shape, dtype=ary.dtype)
```
majosm, line 311: I guess this one might be OK as `ary.dtype`? As long as we assume `f` doesn't convert between scalar types.

majosm, line 311: I guess this one might be OK as `ary.dtype`? As long as we assume `f` doesn't convert to a different scalar type.

grudge/tools.py
```diff
@@ -235,4 +239,105 @@ def build_jacobian(actx, f, base_state, stepsize):
...
+        leaf_cls: Optional[type] = None,
+        return_nested: bool = False) -> np.ndarray:
+    r"""
+    Like :func:`map_subarrays`, but with support for
+    :class:`arraycontext.ArrayContainer`\ s.
+
+    :param leaf_cls: An array container of this type will be considered a leaf
+        and will be passed to *f* without further destructuring.
+    """
+    if type(ary) is leaf_cls or isinstance(ary, np.ndarray):
```
majosm, line 331: I don't think the `isinstance(ary, np.ndarray)` here is specific enough. It should only call `map_subarrays` if `ary` is an array of scalars. For example, say you have the gradient of a vector of `DOFArray`s that was created using `return_nested=True`; it should look like:

```python
ndarray([
    ndarray([DOFArray, DOFArray, DOFArray]),
    ndarray([DOFArray, DOFArray, DOFArray]),
    ndarray([DOFArray, DOFArray, DOFArray])])
```

If you then take the divergence of that, `rec_map_subarrays` should recurse once on the outer `ndarray` and then pass each of the subarrays to `map_subarrays`. But with this check I think it will pass the whole thing to `map_subarrays`.

majosm, line 331: I don't think the `isinstance(ary, np.ndarray)` here is specific enough. It should only call `map_subarrays` if `ary` is an array of scalars. For example, say you have the gradient of a vector of `DOFArray`s that was created using `return_nested=True`; it should look like:

```python
ndarray([
    ndarray([DOFArray, DOFArray, DOFArray]),
    ndarray([DOFArray, DOFArray, DOFArray]),
    ndarray([DOFArray, DOFArray, DOFArray])])
```

If you then take the divergence of that, `rec_map_subarrays` should recurse once on the outer `ndarray` and then pass each of the subarrays to `map_subarrays`. But with this check I think it will pass the whole thing to `map_subarrays`.

inducer (commented):
Thanks for sifting through this! I think what we need here is to judiciously revert some of my changes that ended up breaking stuff for your use case, while keeping/improving some things that (maybe) added value, such as the type annotations.

Could you do that, and add some tests? I'd be OK with merging that once that's done.

grudge/tools.py
majosm: I think it just needs to be `Any`, since the scalar types can be arbitrary.

grudge/tools.py
```diff
@@ -235,4 +239,105 @@ def build_jacobian(actx, f, base_state, stepsize):
...
+                    result[idx + out_slice] = f(ary[idx + in_slice])
+                return result
+
+
+def rec_map_subarrays(
+        f: Callable[[np.ndarray], np.ndarray],
+        in_shape: Tuple[int, ...],
+        out_shape: Tuple[int, ...],
+        ary: np.ndarray, *,
+        leaf_cls: Optional[type] = None,
```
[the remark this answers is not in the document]
majosm: I changed it to assume internally that types for which `np.isscalar(...)` returns `True` are scalars, and made the function argument `scalar_cls`.

grudge/tools.py
majosm: I made it deduce the `dtype` from the function application results.

grudge/tools.py
majosm: Same for this case.

grudge/tools.py
majosm: Fixed it.

majosm:
I partially reverted the changes, added tests, and also added code to deduce the result scalar type. Should be ready for another look

inducer:
This looks lovely! Thanks for working on this.

PR #236, opened 2022-03-17
Support multiple volumes in a discretization collection

### LINKED ISSUE #104: DiscretizationCollection constructor should do no work
[This](https://github.com/inducer/grudge/blob/0a7b2fb2e4e170809b55e0c3c193ae1dbc4a61a4/grudge/discretization.py#L74-L162) is too fat. The constructor itself should just be an attribute setter. The actual work should be done by a "maker function", like `make_discretization_collection`.

inducer:
This isn't finished yet. Missing:

- [x] Finish eager cross-rank inter-volume communication
- [x] Finish lazy cross-rank inter-volume communication
- [ ] Document `*inter_volume_trace_pairs`
- [ ] Test inter-volume communication (i.e. complete the test stub)
- [ ] Test MPI inter-volume communication
- [ ] Needs https://github.com/inducer/meshmode/pull/307
- [x] Needs https://github.com/inducer/arraycontext/pull/150
- [ ] Point `requirements.txt` back at meshmode `main`

cc @majosm 

Closes #104.


inducer:
@majosm I think this is headed towards being not totally broken. Could you take a look to see how things look to you and look into the remaining issues above?

inducer:
The changes discussed in https://github.com/inducer/meshmode/pull/308#issuecomment-1071180522 are now in. Ready to take a look, @majosm.

inducer:
I agree that mutable objects should use the first form. The other aspect that I run into is that if there's multiple layers of functions calling each other, and only the "innermost" function looks at the argument, then I find it cleaner to use `None` instead of redundantly specifying the default. If you're not in either of those cases, I'm not sure I have a preference.

majosm:
Superseded by #280 and #285.

PR #239, opened 2022-03-21
Multiple volumes using partition-ID-enabled meshmode

majosm:
Modifies #236 to use the changes from inducer/meshmode#308.


grudge/discretization.py
```diff
@@ -127,12 +125,38 @@ def _normalize_discr_tag_to_group_factory(
 # }}}
 
 
+# {{{ _PartitionIDHelper
+
+@dataclass(frozen=True, init=True, eq=True)
+class _PartitionIDHelper:
```
inducer, line 131: Let's just have a `PartID` with volume tag and rank?

majosm:
@inducer I think this is ready for a look (in case you happen to have a chance to look through it; no rush).

majosm:
Superseded by #280 and #285.

PR #243, opened 2022-03-25
Attempt to fix inter-volume trace pair functions

majosm:
I abandoned the previous one-way communication idea (I didn't know how to make this work in lazy, since there wouldn't be anything to staple the sends to). This PR is an attempt to implement a two-way version. Still untested.

Since the trace data is possibly heterogeneous in the multi-volume case, the sends in `_RankBoundaryCommunicationLazy` can't be stapled to the receives anymore (since the sent and received data on a given rank will in general have different numbers of subarrays). As a workaround, I staple the sends to arrays of zeros with the same shape as the data being sent. Then I reduce the resulting array container of zeros to a scalar zero value, and add it to the trace pair data being returned so that something (hopefully) will depend on it.

Depends on #239 (and inducer/meshmode#308).


grudge/trace_pair.py
```diff
@@ -572,61 +658,110 @@ class _RankBoundaryCommunicationLazy:
...
-                        src_rank=remote_rank, comm_tag=ary_tag,
-                        shape=local_bdry_subary.shape,
-                        dtype=local_bdry_subary.dtype))
+        # Staple the sends to a bunch of dummy arrays of zeros
+        def send_single_array(key, local_subary):
+            if isinstance(local_subary, Number):
+                return 0
+            else:
+                ary_tag = (comm_tag, key)
+                return staple_distributed_send(
```
inducer, line 695: Make the send separately from stapling it.

majosm:
@inducer This is probably ready for a look too (though it still contains all of #239; I'll leave it as a draft until that gets merged.)

majosm:
Superseded by #280 and #285.

PR #244, opened 2022-03-25
Fix `test_func_comparison_mpi`: actually test communicated value, fix lazy

inducer:


test/test_mpi_communication.py
```diff
@@ -150,19 +150,21 @@ def _test_func_comparison_mpi_communication_entrypoint(actx):
...
-              for tpair in op.cross_rank_trace_pairs(dcoll, myfunc,
-                  comm_tag=SimpleTag))
-    ) - (all_faces_func - bdry_faces_func)
+    def hopefully_zero():
+        return (
+            op.project(
+                dcoll, "int_faces", "all_faces",
+                dcoll.opposite_face_connection()(int_faces_func)
+            )
+            + sum(op.project(dcoll, tpair.dd, "all_faces", tpair.ext)
```
inducer, line 159: This is where the bug was: it used to say `tpair.int`.

majosm (approved):

PR #249, opened 2022-04-06
Bump the right tolerance (not the wrong one) in test_mass_mat_trig

inducer:
Fixes https://github.com/inducer/grudge/pull/248/


test/test_grudge_sym_old.py
```diff
@@ -144,11 +144,11 @@ def _get_variables_on(dd):
 
     num_integral_1 = np.dot(ones_volm, actx.to_numpy(flatten(mass_op(f=f_quad))))
     err_1 = abs(num_integral_1 - true_integral)
-    assert err_1 < 2e-9, err_1
+    assert err_1 < 6e-9, err_1
```
matthiasdiener, line 147:
```suggestion
    assert err_1 < 9e-9, err_1
```

Even with this PR, this test fails on my Mac M1. Could we bump it to 9e-9?

inducer: #250

PR #252, opened 2022-04-12
Add a helper to enable use of compiled LSRK45

inducer:
This seems to work, and it seems to help a fair bit for the wave operator. It's at 0.008s on my laptop. (Earlier measurements were 0.06s, but I forgot what those were.)

Draft because:
- [x] Needs https://github.com/inducer/pytato/pull/308
- [x] ~~Find a way to get rid of the extra `_cls_has_array_context_attr=True`.~~ Can't, currently. See discussion in https://github.com/inducer/arraycontext/pull/155.
- [x] Wait for https://github.com/inducer/arraycontext/pull/155 to make this easier to use.
- [x] Use this in a convergence test.

cc @MTCam

### DIFF
--- a/examples/wave/wave-op-mpi.py
+++ b/examples/wave/wave-op-mpi.py
@@ -30,7 +30,7 @@ import pyopencl as cl
 import pyopencl.tools as cl_tools
 
 from arraycontext import (
-    thaw, freeze,
+    thaw,
     with_container_arithmetic,
     dataclass_array_container
 )
@@ -45,7 +45,7 @@ from meshmode.mesh import BTAG_ALL, BTAG_NONE  # noqa
 from grudge.dof_desc import as_dofdesc, DOFDesc, DISCR_TAG_BASE, DISCR_TAG_QUAD
 from grudge.trace_pair import TracePair
 from grudge.discretization import DiscretizationCollection
-from grudge.shortcuts import make_visualizer, rk4_step
+from grudge.shortcuts import make_visualizer, compiled_lsrk45_step
 
 import grudge.op as op
 
@@ -57,7 +57,8 @@ from mpi4py import MPI
 
 # {{{ wave equation bits
 
-@with_container_arithmetic(bcast_obj_array=True, rel_comparison=True)
+@with_container_arithmetic(bcast_obj_array=True, rel_comparison=True,
+        _cls_has_array_context_attr=True)
 @dataclass_array_container
 @dataclass(frozen=True)
 class WaveState:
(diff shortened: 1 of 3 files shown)


inducer:
This currently rolls over when trying to use the fusion array contractor with
```
NotImplementedError: Cannot fit loop nest 'frozenset()' into known set of loop-nest patterns.
```
cc @kaushikcfd

kaushikcfd:
A temporary workaround for that would be:
```patch
diff --git a/meshmode/array_context.py b/meshmode/array_context.py
index e155671..e9fa6e0 100644
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -1272,6 +1272,19 @@ class FusionContractorArrayContext(
 
         # }}}
 
+        # {{{ un-materialize scalars
+
+        def unmaterialize_scalars(expr):
+            if isinstance(expr, pt.Array) and expr.shape == ():
+                return expr.without_tags(pt.tags.ImplStored(),
+                                         verify_existence=False)
+            else:
+                return expr
+
+        dag = pt.transform.map_and_copy(dag, unmaterialize_scalars)
+
+        # }}}
+
         # {{{ infer axis types
 
         from meshmode.pytato_utils import (unify_discretization_entity_tags,
```

This blatant un-materialization isn't great. But I can't think of a general way to handle this pattern.

inducer:
```
residual = 0*y
```
Applied.

inducer:
There's still a failure here when running distributed:
```
pytools.tag.NonUniqueTagError: Multiple tags are direct subclasses of the following UniqueTag(s): PartIDTag
```
I attribute that to https://github.com/inducer/pytato/issues/307.

examples/wave/wave-op-mpi.py
```diff
@@ -57,7 +57,9 @@
 
 # {{{ wave equation bits
 
-@with_container_arithmetic(bcast_obj_array=True, rel_comparison=True)
+@with_container_arithmetic(bcast_obj_array=True, rel_comparison=True,
+        # FIXME: I'll put together an actx patch to make this unnecessary
+        _cls_has_array_context_attr=True)
```
inducer, line 62: We can delete this once https://github.com/inducer/arraycontext/pull/155 is in.

inducer: Alas, it is not to be. See the discussion in https://github.com/inducer/arraycontext/pull/155.

inducer:
The latest timings here: Without this, 0.0164s per step, with this change 0.0078s per step. There is another usability issue that I'd like to get to before merging this (can't currently pass numpy array scalars to compiled functions). Beyond that, there are quite a few more low-hanging fruit.

inducer:
> There is another usability issue that I'd like to get to before merging this (can't currently pass numpy array scalars to compiled functions).

I've decided that the run time cost of supporting this is too large. We'll have to burden the user with getting rid of array scalars beforehand. (In particular, `numpy.ndarray` satisfies `is_array_container_type`, so "detect array scalars" would be living in a very hot code path.)

PR #255, opened 2022-04-29
 Add EqualityMapper to follow pymbolic

alexfikl:
See inducer/pymbolic#74.

This is only needed in the symbolic parts, which should be deprecated. A good time to rip them out completely?


alexfikl:
Not needed now that the symbolic code is gone :tada:

PR #256, opened 2022-05-17
Make nodal reductions work with empty meshes

majosm:
~~:warning: Depends on inducer/meshmode#330. :warning:~~ (Merged.)

### DIFF
--- a/test/test_reductions.py
+++ b/test/test_reductions.py
@@ -59,7 +59,8 @@ logger = logging.getLogger(__name__)
 @pytest.mark.parametrize(("mesh_size", "with_initial"), [
     (4, False),
     (4, True),
-    (0, True)  # No False case; must pass initial for empty nodal_min/nodal_max
+    (0, False),
+    (0, True)
 ])
 def test_nodal_reductions(actx_factory, mesh_size, with_initial):
     actx = actx_factory()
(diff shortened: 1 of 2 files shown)


test/test_reductions.py
```diff
@@ -82,21 +83,35 @@ def h(x):
...
-    for grudge_op, np_op in [(op.nodal_sum, np.sum),
-                             (op.nodal_max, np.max),
-                             (op.nodal_min, np.min)]:
+    for grudge_op, np_op in [(op.nodal_max, np.max),
+                             (op.nodal_min, np.min),
+                             (op.nodal_sum, np.sum)]:
+        if grudge_op is op.nodal_max:
+            extra_kwargs = {"initial": -100.}
+        elif grudge_op is op.nodal_min:
+            extra_kwargs = {"initial": 100.}
```
inducer, line 92: This should also test the case where `initial` is not passed.

majosm: Hmm, I just noticed something while fixing this: the no-`initial` behavior is currently different between `nodal_min`/`nodal_max` and `numpy.min`/`numpy.max`. For the former I made the default `initial` value +/-`np.inf`, but there is no default `initial` value for the numpy functions (so it fails for zero-size arrays when `initial` is omitted). Should I make `nodal_min`/`nodal_max` behave the same way?

inducer: Yeah. Absent other precedents, I think that's the sane thing to do.

inducer: I'm not sure I totally thought this one through before responding. There is the `numpy` precedent, which guided the local functions:
```
>>> np.min([])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<__array_function__ internals>", line 180, in amin
  File "/home/andreas/src/env-3.10/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 2916, in amin
    return _wrapreduction(a, np.minimum, 'min', axis, None, out,
  File "/home/andreas/src/env-3.10/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 86, in _wrapreduction
    return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
ValueError: zero-size array to reduction operation minimum which has no identity
```
What would blow up if we made the distributed functions consistent with that?

inducer: And sorry about not thinking better sooner! :)

majosm: > What would blow up if we made the distributed functions consistent with that?

Possibly nothing? As I have things right now in https://github.com/inducer/grudge/pull/256/commits/771fb7bab3a18e64248fe7725f53c77a48ef6ae6, `nodal_min_loc`/`nodal_max_loc` raise if they get an empty local mesh and no initial value. In the single-volume case, this shouldn't happen (short of empty test meshes). And in the multi-volume case, the caller just needs to make sure they pass `initial`.

majosm: On the other hand, it's a little annoying to have to specify `initial` everywhere. What do you think about something along the lines of https://github.com/majosm/grudge/commit/fa4016696e571f059e36bd96c61b24ec07af564f? It's a little janky, but seems to work...

inducer: I'm not sure I love that. If there's one thing that I know I'd like to be true, then it's that local and distributed reductions work consistently.

> a little annoying to have to specify initial everywhere.

How many places are we talking?

Compared to https://github.com/majosm/grudge/commit/fa4016696e571f059e36bd96c61b24ec07af564f, I prefer using a default value for `initial` everywhere. The only argument against that is the numpy precedent. OTOH, Loopy already returns the neutral element for empty reductions, and that hasn't bitten us yet. Maybe that's the least bad approach, if specifying `initial` is sufficiently annoying. Sorry about my waffling here! :waffle:

majosm: > I'm not sure I love that. If there's one thing that I know I'd like to be true, then it's that local and distributed reductions work consistently.

FWIW, the implementations in https://github.com/majosm/grudge/commit/fa4016696e571f059e36bd96c61b24ec07af564f are consistent in a certain sense: `initial` is only needed when "the mesh" might be empty. The difference is in which mesh is being considered (global vs. local).

> How many places are we talking?

Not a _huge_ number: maybe a dozen between the prediction driver and mirgecom, a few more in my multi-volume example, and a couple in `dt_utils`.

inducer: Eh, that's enough places to worry about this. My vote would be to have a default value for `initial` in `min` and `max`.

majosm: Reverted.

inducer:
LGTM, thanks!

PR #269, opened 2022-06-22
Fix `build_jacobian`

majosm:
Glanced through `build_jacobian` earlier and I noticed what looks like a bug. It's supposed to be `base_state` here, right?


inducer:
Yikes! Good catch. Thanks for that fix. I pushed a test, too. You never know. :)

PR #273, opened 2022-07-01
Ignore B023 in tests where the pattern it flags is common and innocuous

inducer:


inducer:
(Tried to force merge, but it looks like that's broken on Github at the moment.)

PR #281, opened 2022-09-02
Cleanup in grudge.tools

alexfikl:
Some random changes:

* removes unused `OrderedSet`
* moves some (symbolic?) stuff to `grudge.models.em`, which seems to be the only place using it.
* adds some type annotations to `build_jacobian`
* adds some vim markers here and there


inducer:
Thx!

PR #284, opened 2022-09-19
Rewrite divergence as a single einsum

kaushikcfd:


inducer:
Thanks! This was missing analogous code for the strong-form version. I added that.

PR #246, opened 2022-04-04
Multi-volume extras

majosm:
Adds the remaining missing pieces needed for multi-volume support.

Depends on #243.

PR #285, opened 2022-09-23
Support coupled volumes

majosm:
Adds the rest of the changes from the original #236, #239, #243, #246 chain. Depends on #280.


grudge/discretization.py
```diff
@@ -1015,4 +1017,44 @@ def make_discretization_collection(
 # }}}
 
 
+# {{{ filter_part_boundaries
+
+def filter_part_boundaries(
```
inducer, line 1022: `get_boundary_dofdescs`? ("Filter" is possibly confusing with filtering-for-dissipation. And the function name didn't give me a vibe of returning `DOFDesc`s.)

inducer:
How far away is this from being reviewable?

majosm:
Just need to add docs and figure out how to avoid duplicating the mesh in the part ID normalization step (to avoid https://github.com/inducer/meshmode/issues/358).

PR #292, opened 2022-11-14
Update mirgecom downstream testing to use integrated test runner.

MTCam:
Use mirgecom integrated test runner to run lazy, examples, and production testing.  Plugs a production testing hole.


inducer:
Thanks for this contribution.

Analogous CI code exists in, at least, arraycontext, pytato, and loopy. What's provided here looks a bit "non-final", and I'd prefer not to have to continue making changes in multiple places as things get finalized. Instead, I'd prefer if mirgecom had a script entrypoint for downstream tests, so that upstream package can defer the policy decision of what should constitute downstream CI can be deferred to mirgecom.

PR #297, opened 2022-12-14
Pytato+distributed: Verify DAG before partitioning

matthiasdiener:
Needs 
- https://github.com/inducer/pytato/pull/404


inducer:
This is possibly moot given that the current state of https://github.com/inducer/pytato/pull/393 performs these checks implicitly as the partition is computed.

PR #298, opened 2023-01-10
Support new find_distributed_partition

matthiasdiener:
https://github.com/inducer/pytato/pull/393 changes the function signature. Not sure if this is the best way to do it, but it appears to work.

### DIFF
--- a/grudge/array_context.py
+++ b/grudge/array_context.py
@@ -232,21 +232,21 @@ class _DistributedLazilyPyOpenCLCompilingFunctionCaller(
         self.actx._compile_trace_callback(self.f, "pre_find_distributed_partition",
                 dict_of_named_arrays)
 
-         # https://github.com/inducer/pytato/pull/393 changes the function signature
-         try:
-             # pylint: disable=too-many-function-args
-             distributed_partition = pt.find_distributed_partition(
-                 # pylint-ignore-reason:
-                 # '_BasePytatoArrayContext' has no
-                 # 'mpi_communicator' member
-                 # pylint: disable=no-member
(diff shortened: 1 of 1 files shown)


grudge/array_context.py
```diff
@@ -239,7 +239,20 @@ def _dag_to_compiled_func(self, dict_of_named_arrays,
...
+            # https://github.com/inducer/pytato/pull/393 changes the
+            # function signature
+            # pylint: disable=too-many-function-args
+            distributed_partition = pt.find_distributed_partition(
+                                        # pylint-ignore-reason:
+                                        # '_BasePytatoArrayContext' has no
+                                        # 'mpi_communicator' member
+                                        # pylint: disable=no-member
+                                        self.actx.mpi_communicator,
+                                        dict_of_named_arrays)
```
majosm, line 255:
```suggestion
         # https://github.com/inducer/pytato/pull/393 changes the function signature
         try:
             # pylint: disable=too-many-function-args
             distributed_partition = pt.find_distributed_partition(
                 # pylint-ignore-reason:
                 # '_BasePytatoArrayContext' has no
                 # 'mpi_communicator' member
                 # pylint: disable=no-member
                 self.actx.mpi_communicator, dict_of_named_arrays)
         except TypeError as e:
             if "find_distributed_partition() takes 1 positional" in str(e):
                 distributed_partition = pt.find_distributed_partition(
                     dict_of_named_arrays)
             else:
                 raise
```
(So it doesn't catch every `TypeError`, just the relevant one.)

inducer: Eh, sure. I can live with this.

PR #300, opened 2023-01-19
Make compatible with new distributed DAG partitioner

majosm:


inducer:
This looks similar to #298? I had more or less the same discussion with @matthiasdiener there, and I think my preference would be to just merge https://github.com/inducer/pytato/pull/393/ and #298 at the same time.

majosm:
> This looks similar to #298? I had more or less the same discussion with @matthiasdiener there, and I think my preference would be to just merge [inducer/pytato#393](https://github.com/inducer/pytato/pull/393) and #298 at the same time.

Oops, I missed that one. 🙂

PR #301, opened 2023-01-23
Enforce comm_tag in lazy evaluation

matthiasdiener:


matthiasdiener:
Nvm, this is already done in the constructor

PR #302, opened 2023-01-25
Attach axes tags to cell volume terms in dt_geometric_factors

kaushikcfd:
After doing this the `models.wave` can be digested by [BatchedEinsumArrayContext](https://github.com/inducer/arraycontext/pull/217) without falling back.


inducer:
Thanks!

PR #303, opened 2023-02-02
Optimize maxwell driver by separating real and imaginary components

kaushikcfd:


inducer:
Sorry, not loving this. I'd prefer finishing up #220 instead, i.e. making sure we use real dtypes for the operator matrices.

kaushikcfd:
Fair. Closing this for that.

PR #308, opened 2023-05-03
Use `comm.scatter` for mesh distribution

inducer:
- cf. https://github.com/inducer/meshmode/pull/342.
- goes with (but doesn't require) https://github.com/inducer/meshmode/pull/376.

### DIFF
--- a/examples/wave/wave-min-mpi.py
+++ b/examples/wave/wave-min-mpi.py
@@ -49,7 +49,7 @@ class WaveTag:
 
 def main(ctx_factory, dim=2, order=4, visualize=False):
     comm = MPI.COMM_WORLD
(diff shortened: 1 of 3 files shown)


majosm:
Hmm, I think the mirgecom tests are failing because the wall coupling was merged (and therefore mirgecom now depends on my #285 branch). I'm going to check if there's an easy way to make those changes conditional so the CI still works here in the meantime.

inducer:
No need IMO, as long as #285 goes in soon. I can just ignore the mirgecom failures for the time being. But I've put the downstream issue on the agenda for Friday, too, to at least have a discussion.

majosm (approved):
LGTM.

PR #310, opened 2023-05-30
Fuse Indirection Maps

kaushikcfd:
For wave's lift operator,
- Before:
before_indirections_fusion

- After:
after_indirection_fusion


test/test_pytato_transforms.py
```diff
@@ -0,0 +1,225 @@
...
+    nel_1d = 4
+    mesh = generate_regular_rect_mesh(
+        a=(-0.5,)*dim,
+        b=(0.5,)*dim,
+        nelements_per_axis=(nel_1d,)*dim,
+        boundary_tag_to_face={"bdry": ["-x", "+x",
+                                       "-y", "+y",
+                                       "-z", "+z"]}
+    )
+    dcoll = grudge.make_discretization_collection(ref_actx, mesh, order=2)
```
kaushikcfd, line 210: @inducer: I noticed a subtle and sub-optimal behavior that is worth noting. If we are using PyOpenCLActx as a setup actx and PytatoActx as an operational actx, then the frozen geometric factors (such as normals) would be stored at per-DOF even for affine geometry mappings. Such use-cases will bump memory footprint.

PR #311, opened 2023-05-31
Add axis tags for reshape in `_apply_elementwise_reduction`

majosm:
This is part of what was causing the fallback warnings in the Y3 production case.


inducer:
Thanks!

PR #313, opened 2023-08-01
Tensor product differentiation operators

a-alveyblanc:
This PR seeks to special-case local gradient, divergence, and derivative operators for tensor product finite elements

- [ ] Single axis derivatives?


grudge/op.py
```diff
@@ -166,6 +166,38 @@
     )
 
 
+# {{{ Temporary tools for tensor product operators
+# NOTE: Will possibly be removed in a future version of tensor product operator
+# development since (I think) it is not entirely necessary
+from pytools.tag import Tag
+class OutputIsTensorProductDOFArrayOrdered(Tag):
```
inducer, line 173: Move this to somewhere near the array context.

grudge/op.py
```diff
@@ -166,6 +166,38 @@
...
+                            + ",".join(f"N{i}"
+                                       for i in range(len(arg.shape)-1))
+                            ))
+
+                    new_args.append(arg)
+
+                knl = knl.copy(args=new_args)
+                t_unit = t_unit.with_kernel(knl)
+
+        return super().transform_loopy_program(t_unit)
```
inducer, line 197: This is also going to have a lazy tentacle.

grudge/op.py
```diff
@@ -202,19 +234,97 @@ def _gradient_kernel(actx, out_discr, in_discr, get_diff_mat, inv_jac_mat, vec,
...
+        # apply differentiation matrix to vec
+        # check len(vec.shape) since shape is expected to be
+        # (nelements, nnodes1d, nnodes1d)
+        if len(vec.shape) == 3:
+            specs = ["il,elj->eij",
+                     "jl,eil->eij"]
+        elif len(vec.shape) == 4:
+            specs = ["il,eljk->eijk",
+                     "jl,eilk->eijk",
+                     "kl,eijl->eijk"]
```
inducer, line 264: Can you make this dimension-independent?

grudge/op.py
```diff
@@ -202,19 +234,97 @@ def _gradient_kernel(actx, out_discr, in_discr, get_diff_mat, inv_jac_mat, vec,
...
+        else:
+            specs = None
+        assert specs is not None
```
inducer, line 267: Use `assert` only for invariants that you think you can prove.

grudge/op.py
```diff
@@ -202,19 +234,97 @@ def _gradient_kernel(actx, out_discr, in_discr, get_diff_mat, inv_jac_mat, vec,
...
+
+        diff_mat = get_diff_mat(actx, grp, grp)
+        grad = make_obj_array([
+                actx_tp.einsum(
+                    spec,
+                    diff_mat,
+                    vec,
+                    arg_names=("diff_mat", "vec"),
+                    tagged=(FirstAxisIsElementsTag(),
```
inducer, line 276: `FirstAxisIsElementsTag()` is old-timey, should not use in new code.

grudge/op.py
```diff
@@ -259,16 +369,37 @@ def _reference_derivative_matrices(actx: ArrayContext,
...
+            space1d = grp.space.bases[0]
+            shape1d = grp.shape.bases[0]
+
+            nodes1d = mp.edge_clustered_nodes_for_space(space1d, shape1d)
+            basis1d = mp.basis_for_space(space1d, shape1d)
+
+            vdm1d = mp.vandermonde(basis1d.functions, nodes1d)
+            vdm_p1d = mp.vandermonde(basis1d.gradients, nodes1d)[0]
+
+            return actx.freeze(actx.from_numpy(vdm_p1d @ la.inv(vdm1d)))
```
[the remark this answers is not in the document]
a-alveyblanc: possibly fixed in meshmode [PR 384](https://github.com/inducer/meshmode/pull/384)

grudge/op.py
```diff
@@ -202,19 +206,154 @@ def _gradient_kernel(actx, out_discr, in_discr, get_diff_mat, inv_jac_mat, vec,
         *, metric_in_matvec):
     # See _single_axis_derivative_kernel for comments on the usage scenarios
     # (both strong and weak derivative) and their differences.
+
+    def compute_tensor_product_grad(actx, grp, diff_mat, vec, ijm,
+                                    metric_in_matvec):
```
inducer, line 211: Document what `metric_in_matvec` means.

grudge/op.py
```diff
@@ -202,19 +206,154 @@ def _gradient_kernel(actx, out_discr, in_discr, get_diff_mat, inv_jac_mat, vec,
...
-                    ijm_i,
-                    get_diff_mat(
-                        actx,
-                        out_element_group=out_grp,
-                        in_element_group=in_grp
-                    ),
-                    vec_i,
-                    arg_names=("inv_jac_t", "ref_stiffT_mat", "vec"),
-                    tagged=(FirstAxisIsElementsTag(),))
+        else actx.einsum(
```
inducer, line 345: Factor this into a function also.

requirements.txt
```diff
@@ -10,7 +10,7 @@ git+https://github.com/inducer/leap.git#egg=leap
 git+https://github.com/inducer/meshpy.git#egg=meshpy
 git+https://github.com/inducer/modepy.git#egg=modepy
 git+https://github.com/inducer/arraycontext.git#egg=arraycontext
-git+https://github.com/inducer/meshmode.git#egg=meshmode
+git+https://github.com/a-alveyblanc/meshmode.git@tensor-product-1d-nodes-and-1d-basis#egg=meshmode
```
inducer, line 13: Back out

test/test_op.py
```diff
@@ -159,6 +160,140 @@ def get_flux(u_tpair):
...
+
+@pytest.mark.parametrize("form", ["strong", "weak"])
+@pytest.mark.parametrize("dim", [2, 3])
+@pytest.mark.parametrize("order", [2, 3])
+@pytest.mark.parametrize(("vectorize", "nested"), [
+    (False, False),
+    (True, False),
+    (True, True)
+    ])
+def test_tensor_product_gradient(form, dim, order, vectorize,
```
inducer, line 172: Do this by parametrizing the original test?

test/test_op.py
```diff
@@ -275,6 +410,135 @@ def get_flux(u_tpair):
...
+
+@pytest.mark.parametrize("form", ["strong", "weak"])
+@pytest.mark.parametrize("dim", [2, 3])
+@pytest.mark.parametrize("order", [2, 3])
+@pytest.mark.parametrize(("vectorize", "nested"), [
+    (False, False),
+    (True, False),
+    (True, True)
+    ])
+def test_tensor_product_divergence(form, dim, order, vectorize,
```
inducer, line 422: Parametrize original?

grudge/op.py
```diff
@@ -179,77 +197,353 @@ def _single_axis_derivative_kernel(
...
 
 def _divergence_kernel(actx, out_discr, in_discr, get_diff_mat, inv_jac_mat, vec,
         *, metric_in_matvec):
     # See _single_axis_derivative_kernel for comments on the usage scenarios
     # (both strong and weak derivative) and their differences.
+
+
+    # {{{ tensor product div
+
+    def compute_tensor_product_div(actx, grp, diff_mat, vec, ijm):
```
MTCam, line 430: The result of the simplicial version of these operators lives on the the base discretization.  The TPE versions produce results that live on the quadrature discretization, preventing overintegration from working as expected.

a-alveyblanc: Are you referring to the use of `in_grp` and `out_grp` between the two versions?

MTCam: > Are you referring to the use of `in_grp` and `out_grp` between the two versions?

That's definitely a code difference I can see, but I'm actually referring to the observation that if you send in *vec* that is on the quadrature discretization, this routine produces a result that lives on the quadrature discretization instead of one that lives on the base discretization as the simplicial version does.

a-alveyblanc: >if you send in vec that is on the quadrature discretization, this routine >produces a result that lives on the quadrature discretization instead of one >that lives on the base discretization

I think the reason this happens is because the routine only considers `in_grp`. I'm not sure why I wrote this to only consider `in_grp`. I *think* I remember some odd behavior (or outright errors) when grabbing the 1D operators.

MTCam: There seems to be only a TPE case in this routine for when in_group == out_group:

https://github.com/a-alveyblanc/grudge/blob/a662445c43d38d97324cb9744d144e0fa9a47a42/grudge/op.py#L755

a-alveyblanc: I remember writing it that way, but I can't remember the exact reason. I'll look into it a bit and report back

PR #316, opened 2023-09-06
Use `make_dict_of_named_arrays` to suppress `tags=None` deprecation warning

majosm:


inducer (approved):

inducer:
The documentation failure is also on main, and I haven't yet had a chance to dig, so don't let that keep you from un-drafting.

PR #319, opened 2023-09-13
Add a CommTag class with a stable hash

matthiasdiener:
TODO:

- [x] add tests

### DIFF
--- a/grudge/trace_pair.py
+++ b/grudge/trace_pair.py
@@ -334,6 +334,9 @@ class CommTag:
         key_builder.rec(key_hash, (self.__class__.__module__,
                                    self.__class__.__qualname__))
 
+    def __repr__(self) -> str:
+        return self.__class__.__module__ + "." + self.__class__.__qualname__
+
 
 def interior_trace_pairs(dcoll: DiscretizationCollection, vec, *,
         comm_tag: Optional[Hashable] = None, tag: Hashable = None,
--- a/grudge/trace_pair.py
+++ b/grudge/trace_pair.py
@@ -327,9 +327,6 @@ class CommTag:
     def __hash__(self):
         return hash(tuple(str(type(self)).encode("ascii")))
 
-    def __eq__(self, other):
-        return isinstance(other, type(self))
-
 
 def interior_trace_pairs(dcoll: DiscretizationCollection, vec, *,
         comm_tag: Optional[Hashable] = None, tag: Hashable = None,
--- a/grudge/trace_pair.py
+++ b/grudge/trace_pair.py
@@ -6,6 +6,7 @@ Container class and auxiliary functionality
 -------------------------------------------
 
 .. autoclass:: TracePair
+.. autoclass:: CommTag
 
 .. currentmodule:: grudge.op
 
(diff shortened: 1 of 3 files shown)


matthiasdiener:
This is still a draft, but it's ready for a first look @inducer @majosm

grudge/trace_pair.py
```diff
@@ -318,8 +319,20 @@ def interior_trace_pair(dcoll: DiscretizationCollection, vec) -> TracePair:
...
+    def __hash__(self):
+        return hash(tuple(str(type(self)).encode("ascii")))
+
+    def __eq__(self, other):
+        return isinstance(other, type(self))
+
+
 def interior_trace_pairs(dcoll: DiscretizationCollection, vec, *,
-        comm_tag: Hashable = None, tag: Hashable = None,
+        comm_tag: CommTag = None, tag: Hashable = None,
```
inducer, line 335: I don't agree with this change. There's  no good reason why other `Hashable`s should not be accepted.

grudge/trace_pair.py
```diff
@@ -318,8 +319,20 @@ def interior_trace_pair(dcoll: DiscretizationCollection, vec) -> TracePair:
     return local_interior_trace_pair(dcoll, vec)
 
 
+@dataclass(frozen=True)  # for KeyBuilder support
+class CommTag:
```
inducer, line 323: It doesn't really make sense for all `CommTag`s to be subclasses of this. Some may have data contained in them.

grudge/trace_pair.py
```diff
@@ -318,8 +319,20 @@ def interior_trace_pair(dcoll: DiscretizationCollection, vec) -> TracePair:
...
+    """A communication tag with a hash value that is stable across
+    runs, even without setting ``PYTHONHASHSEED``."""
+    @memoize_method
+    def __hash__(self):
+        return hash(tuple(str(type(self)).encode("ascii")))
+
+    def __eq__(self, other):
+        return isinstance(other, type(self))
```
inducer, line 331: This is incorrect: It accepts arbitrary subclasses for `other`, and violates the property of "`__eq__` implies hash equality."

grudge/trace_pair.py
matthiasdiener: I undid this change in [9b2ced4](https://github.com/inducer/grudge/pull/319/commits/9b2ced47c82129bfdc381292cca900ba0da2d240). I left the warning in https://github.com/inducer/grudge/pull/319/files#diff-5fa6e6b713d37028c16f1bbc5a6d0b4547ecba0f86c8a52eb30e1ba1dc2614e4R518 in  place for now, would you like me to remove it?

grudge/trace_pair.py
matthiasdiener: I added a small test in [9a63df4](https://github.com/inducer/grudge/pull/319/commits/9a63df4e774e517918ae944dbc89e8b2bac17568) with subclassed dataclasses, is that what you had in mind?

grudge/trace_pair.py
matthiasdiener: This is (hopefully) fixed in [6918c63](https://github.com/inducer/grudge/pull/319/commits/6918c63ca6a69829ed9bfb805045657020a2187e)

grudge/trace_pair.py
matthiasdiener: Ok, would you like to keep the `CommTag` class as an optional way to construct the comm tags? i removed the warning in 945406730bd968fdde9688dbb7ce604a329b7110.

matthiasdiener:
It's unlikely this will be needed for deterministic execution, closing.

PR #326, opened 2024-02-05
dt_geometric_factors for quads/hexes

MTCam:
Adds dt_geometric_factors for quads/hexes and extends existing tests for that and RK4 DT estimate to include quads/hexes.


inducer:
Superseded by #338, since I can't seem to push here.

PR #329, opened 2024-02-23
`test_gradient`: test warped mesh, overintegration

inducer:
- [x] Drop meshmode-repoint

### DIFF
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -389,15 +389,15 @@ class DiscretizationCollection:
         base_group_factory = self.group_factory_for_discretization_tag(
                 dd.discretization_tag)
 
-        def geo_group_factory(megrp, index):
+        def geo_group_factory(megrp):
             from modepy.shapes import Simplex
             from meshmode.discretization.poly_element import \
                     PolynomialEquidistantSimplexElementGroup
             if megrp.is_affine and issubclass(megrp._modepy_shape_cls, Simplex):
                 return PolynomialEquidistantSimplexElementGroup(
-                        megrp, order=0, index=index)
+                        megrp, order=0)
             else:
-                return base_group_factory(megrp, index)
+                return base_group_factory(megrp)
 
         from meshmode.discretization import Discretization
         geo_deriv_discr = Discretization(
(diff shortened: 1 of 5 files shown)


grudge/discretization.py
```diff
@@ -107,11 +108,8 @@ def _normalize_discr_tag_to_group_factory(
     assert discr_tag_to_group_factory is not None
 
     # Modal discr should always come from the base discretization
-    if DISCR_TAG_MODAL not in discr_tag_to_group_factory:
-        discr_tag_to_group_factory[DISCR_TAG_MODAL] = \
-            _generate_modal_group_factory(
-                discr_tag_to_group_factory[DISCR_TAG_BASE]
-            )
+    if DISCR_TAG_MODAL not in discr_tag_to_group_factory and order is not None:
+        discr_tag_to_group_factory[DISCR_TAG_MODAL] = ModalGroupFactory(order)
```
majosm, line 112: @inducer Question about this: shouldn't the modal group factory be set up in the `order is None` case too? Can we retrieve the order from whatever factory is being used for `DISCR_TAG_BASE`? (Seeing some failures related to this in mirgecom, hence my asking.)

inducer: I agree, it would be ideal if the modal group was set up unconditionally. Unfortunately, I couldn't figure out a robust way to make that happen. Group factories are just callables, so `discr_tag_to_group_factory[DISCR_TAG_BASE].order` isn't guaranteed to work. We could try that (and give up if it doesn't work), which would make the situation incrementally better. (The previous version of the code unconditionally did that, so...) But it's not a bulletproof fix either. I would be OK with a PR that does this.

majosm: Would it be better to just set it up on the mirgecom side inside [create_discretization_collection](https://github.com/illinois-ceesd/mirgecom/blob/9d4f9a8aa152497d70a75bf79a5d863663d20b93/mirgecom/discretization.py#L41)? We have access to `order` there.

majosm: (Something like https://github.com/illinois-ceesd/mirgecom/pull/1013.)

inducer: Yes, that's a good approach.

PR #331, opened 2024-02-27
Make `FusionContractorArrayContext` fallbacks optional (disabled by default)

majosm:
Updates `MPIPytatoArrayContextBase.__init__` to pass along the arguments added to `FusionContractorArrayContext` to override the new default fallback behavior (raising an exception).

### DIFF
--- a/grudge/array_context.py
+++ b/grudge/array_context.py
@@ -384,9 +384,11 @@ class _DistributedCompiledFunction:
 
 class MPIPytatoArrayContextBase(MPIBasedArrayContext):
     def __init__(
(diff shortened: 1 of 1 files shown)


majosm:
Hrm, I don't think this is going to work as-is. Looks like `MPIPytatoArrayContextBase` needs to be compatible with both `FusionContractorArrayContext` and `PytatoPyOpenCLArrayContext`, which now have different `__init__` arguments.

@inducer, @matthiasdiener Any thoughts on how to refactor this to make it work? Multiple inheritance in Python is not my strong suit...

PR #332, opened 2024-03-06
discretization: use part_id instead of removed part_nr

alexfikl:
Missed this when working on inducer/meshmode#396.

### DIFF
--- a/grudge/discretization.py
+++ b/grudge/discretization.py
@@ -1,5 +1,4 @@
 """
-
 .. autoclass:: DiscretizationTag
 
 .. currentmodule:: grudge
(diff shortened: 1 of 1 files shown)


inducer:
Thx!

inducer:
Out of curiosity: how come the downstream test didn't flag this?

alexfikl:
> Out of curiosity: how come the downstream test didn't flag this?

Hm, seems to be in `distributed_boundary_swap_connection`.. Do the tests not call that or does the upstream run not use MPI?

inducer:
MPI gets turned off for downstream: https://github.com/inducer/ci-support/blob/f2a4fff5ebfb7e37b0eb1e40371471d76f084ced/ci-support.sh#L818

inducer:
So no mystery here. Thanks for looking... I feel like I could've answered that myself. :grimacing: Sorry!

PR #341, opened 2024-05-21
Fix _to_identifier

matthiasdiener:
Fixes https://github.com/inducer/arraycontext/issues/260.

I think running `isidentifier` on individual characters doesn't make too much sense; there are other occurrences of this in the code:

- https://github.com/inducer/grudge/blob/5b73a2b640458b0b08e7d2b4fc41743c1f18c2fa/grudge/dof_desc.py#L91-L95

- https://github.com/inducer/arraycontext/blob/146c598c688b519eb4cc6be5966a56e2a6e6292e/arraycontext/impl/pytato/compile.py#L55-L56

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ def main():
         python_requires="~=3.8",
         install_requires=[
             "pytest>=2.3",
-            "pytools>=2020.3",
+            "pytools>=2024.1.3",
             "modepy>=2013.3",
             "arraycontext>=2021.1",
             "meshmode>=2020.2",


grudge/array_context.py
```diff
@@ -163,7 +163,8 @@ class MPIBasedArrayContext:
 # {{{ distributed + pytato
 
 def _to_identifier(s: str) -> str:
-    return "".join(ch for ch in s if ch.isidentifier())
+    # Only allow digits, letters, and underscores in identifiers
+    return "".join(ch for ch in s if ch.isalnum() or ch == "_")
```
inducer, line 167: Huh. I would've expected these two bits of code to be identical. Are they not?

matthiasdiener: They are not identical, a Python identifier can not start with a number (see e.g. https://docs.python.org/3/reference/lexical_analysis.html#identifiers).

```
>>> "1".isidentifier()
False
>>> "1".isalnum()
True
>>> def _to_identifier(s: str) -> str:
...     # Only allow digits, letters, and underscores
...     return "".join(ch for ch in s if ch.isalnum() or ch == "_")
...
>>> _to_identifier("123")
'123'
>>> def _to_identifier_old(s: str) -> str:
...     return "".join(ch for ch in s if ch.isidentifier())
...
>>> _to_identifier_old("123")
''
```

matthiasdiener: (see also https://github.com/inducer/pytools/pull/225 for a different approach)

inducer: Ah! :facepalm: Makes sense, thanks!

And I think your proposed approach of centralizing this functionality in pytools is reasonable (as you probably guessed from my merging that). I'll roll a pytools release so you can start relying on this.

matthiasdiener: Sounds good! Changed in [3c36e09](https://github.com/inducer/grudge/pull/341/commits/3c36e09b02b84b035ef254b719549975c604237c)

inducer:
https://pypi.org/project/pytools/2024.1.3/

PR #353, opened 2024-07-12
Remove deprecations

alexfikl:
There's quite a lot of them in the latest version. 

This was mainly motivated by porting from the meshmode `flatten` to the arraycontext one, so that we can remove all the container traversal stuff from there, but went on and cleaned some other things up too.

Probably best viewed commit-by-commit, since some of them are straightforward.


grudge/models/euler.py
```diff
@@ -76,7 +79,11 @@ class ConservedEulerField:
 
     @property
     def array_context(self):
-        return self.mass.array_context
+        return (
+            self.mass.array_context
+            if isinstance(self.mass, DOFArray)
+            # NOTE: ConservedEulerField also holds the fluxes sometimes...
+            else self.mass[0].array_context)
```
alexfikl, line 86: I'm not sure what the status quo is on this, where the variables and the flux share the same container. Is it ok to get the array context like this?

grudge/models/em.py
```diff
@@ -220,10 +220,12 @@ def __init__(self, dcoll, epsilon, mu,
...
-        self.pec_tag = pec_tag
-        self.pmc_tag = pmc_tag
-        self.absorb_tag = absorb_tag
-        self.incident_tag = incident_tag
+        from grudge.dof_desc import as_dofdesc
+
+        self.pec_tag = as_dofdesc(pec_tag)
+        self.pmc_tag = as_dofdesc(pmc_tag)
+        self.absorb_tag = as_dofdesc(absorb_tag)
+        self.incident_tag = as_dofdesc(incident_tag)
```
alexfikl, line 228: Note to self: may need to revert this since a tag is not a dofdesc :thinking:

grudge/array_context.py
```diff
@@ -509,11 +510,34 @@ class PytestPyOpenCLArrayContextFactory(
...
+        if actx.allocator is not None:
+            return actx
+
+        from pyopencl.tools import ImmediateAllocator, MemoryPool
+        alloc = MemoryPool(ImmediateAllocator(actx.queue))
+
+        return self.actx_class(
+                actx.queue,
+                allocator=alloc,
+                force_device_scalars=self.force_device_scalars)
```
alexfikl, line 524: Added a memory pool to the tests, since the grudge `PyOpenCLArrayContext` recommended it. Not sure if this is a good idea?

alexfikl:
@inducer This should be good for a look now. I left two questions above, but otherwise it seems ok. The remaining (7k-ish) warnings are about
* Some SVM `InconsistentOpenCLQueueWarning` warnings.
* Some warnings about the number of arguments in various kernels.
* Warnings about some unevaluated arguments to functions.
* Some unused inames.
* Some warnings about accessing `Discretization.mpi_communicator`.

Not quite sure what to do about those, since they're not simple ports, so left them alone :grin:

grudge/models/euler.py
inducer: Well, I definitely do not love the way this looks. :slightly_smiling_face: 

The root cause of the mess, I think, is that we haven't decided the fate of https://github.com/inducer/arraycontext/issues/162.

IMO, this can go one of two ways that are somewhat sane:

- Either: We say "array containers do not know their array context", this problem here goes away, `DOFArray.array_context` disappears, and we have to contend with large-scale interface breakage, because many places rely on fishing array contexts out of `DOFArray`s.
- Or: we double down on array containers knowing their array context, in which case this struct should receive its own stored `array_context` field that is set by `freeze` and `thaw`. This depends on registration with `with_array_context`, which could conceivably be automated by `with_container_arithmetic`. Note that this is all quite inefficient: For each `freeze` and `thaw`, we traverse the whole container hierarchy twice. Once to do the freeze/thaw, and once again to update all the buried array contexts.

I'm not loving our choices here, but I don't think that continuing without picking one of the two is a great option. We'll just get more mess like the above. I'm somewhat leaning towards the first. How do you all feel about this?

cc @majosm @kaushikcfd @MTCam

inducer: FWIW, for this particular instance, I've pushed a change that switches this to explicit-actx: 587690a7d83ec10630ed9394291d7b69f36ae18e. This generally feels like the right thing to do, IMO.

alexfikl: Hm, this seems like a bigger change and I'm more than happy to revert that commit for this PR.

That said, I'm also mostly in favor of the first option, at least for the case where the callee needs an array context and is forced to fish one out of whatever container it received.

However, how is the first option supposed to work with arithmetic? From what I recall that `_cls_has_array_context_attr` thing was introduced so that the arithmetic knows the array types of the context and can broadcast across object arrays and whatnot.

grudge/op.py
```diff
@@ -439,19 +439,19 @@ def _reference_stiffness_transpose_matrices(
...
-
-            mmat = mass_matrix(out_grp)
+            mmat = mp.mass_matrix(out_grp.basis_obj(), out_grp.unit_nodes)
+            diff_matrices = mp.diff_matrices(out_grp.basis_obj(), out_grp.unit_nodes)
             return actx.freeze(
                 actx.tag_axis(1, DiscretizationDOFAxisTag(),
                     actx.from_numpy(
                         np.asarray(
-                            [dmat.T @ mmat.T for dmat in diff_matrices(out_grp)]))))
+                            [dmat.T @ mmat.T for dmat in diff_matrices]))))
```
alexfikl, line 448: I mostly left these alone because the ones in meshmode are cached on the group. It's probably not a big issue here, since the whole function gets memoized anyway.

inducer: Ah, thanks for pointing out the memoization angle. That said, in my local changes to meshmode, I've deleted those two meshmode functions, so... :slightly_smiling_face:

inducer: At any rate, I agree, since grudge memoizes, no issue here.

grudge/models/euler.py
alexfikl: > The more we can simplify the logic there, the better.

Hard agree on that! That chunk of code has been black box of spaghetti to me :cry:

majosm: I'm possibly misunderstanding, but isn't this what `get_container_context_recursively` is for? Does the extra traversal really contribute much time? Seems like it would be negligible compared to all of the other processing going on.

alexfikl:
@inducer The examples run on the CI seems to be taking about ~20min more than before (mostly on the Euler vortex). It doesn't seem to appreciate the changes for some reason..

Previous run on this PR:
https://github.com/inducer/grudge/actions/runs/9925754509/job/27491794872?pr=353

grudge/models/euler.py
inducer: It's not that `get_container_context_recursively` is necessarily costly time-wise (I don't know---it might be?). It's that it can fail: All items in an array container can be, more or less, anything, including scalars. As a result, it can always fail. As a result, any code relying on it is not much better than the spaghetti above, at least IMO.

MTCam: > It's not that `get_container_context_recursively` is necessarily costly time-wise (I don't know---it might be?). It's that it can fail: All items in an array container can be, more or less, anything, including scalars. As a result, it can always fail. As a result, any code relying on it is not much better than the spaghetti above, at least IMO.

Failure =  `get_container_context_recursively` returns `None`?  I guess I might be the only fool waving a flag that says he prefers the 2nd option; double down.

grudge/array_context.py
inducer: That should be OK, as long as that array context gets freed.

inducer:
Thanks for pointing that out. That "previous run" was 33m, the [most recent time on main](https://github.com/inducer/grudge/actions/runs/9962052200/job/27524962185) was 35m... optimistically, this is just noise.

Issue #173, opened 2021-10-12
WADG should use overintegration


inducer:
@thomasgibson pointed out that our WADG could/should be able to take advantage of overintegration, which it currently doesn't. Consider the matrix we're supposed to apply:

![image](https://user-images.githubusercontent.com/352067/137021698-ec2221fc-c9d6-4c3a-8e3b-e4757c3560fe.png)

(From [the paper](https://arxiv.org/pdf/1608.03836.pdf#page=6)) Consider the current implementation:

https://github.com/inducer/grudge/blob/2af3528391bf4cb07f5aacf43f7c6afcc18f693a/grudge/op.py#L645-L653

The `ref_Minv * ref_M * (1/jac_det) * ref_Minv * u` shold really be (in weird pseudocode):

```
ref_Minv * ref_M("overintegrated") * (1/project("overintegrated", jac_det) * project("overintegrated", ref_Minv * u)
```

cc @lukeolson

PR #354, opened 2024-07-19
Use overintegration in WADG

MTCam:
This change set is intended to update WADG to use overintegration (#173).

### DIFF
--- a/grudge/op.py
+++ b/grudge/op.py
@@ -93,6 +93,7 @@ from grudge.discretization import DiscretizationCollection
 from grudge.dof_desc import (
     DD_VOLUME_ALL,
     DISCR_TAG_BASE,
+    DISCR_TAG_QUAD,
     FACE_RESTR_ALL,
     DOFDesc,
     VolumeDomainTag,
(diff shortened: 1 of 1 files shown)


grudge/op.py
```diff
@@ -819,6 +820,96 @@ def _apply_inverse_mass_operator(
...
+    # true_Minv ~ ref_Minv * ref_M * (1/jac_det) * ref_Minv
+    # Overintegration version of action on *vec*:
+    # true_Minv ~ ref_Minv * P(ref_M) * 1/P(Jac) * P(Minv*vec)
+    # P => projection to quadrature
+
+    # Compute 1/P(Jac)
+    inv_area_elements = 1/project(
+        dcoll, dd_base, dd_quad, area_element(
+            actx, dcoll, dd=dd_base,
+            _use_geoderiv_connection=actx.supports_nonscalar_broadcasting))
```
inducer, line 855: The area element is already nonlinear. Its computation should take place on the target discretization.

grudge/op.py
```diff
@@ -819,6 +820,96 @@ def _apply_inverse_mass_operator(
...
+
+    # Compute Minv*vec
+    def apply_minv_to_vec(vec, ref_inv_mass):
+        return actx.einsum(
+            "ij,ej->ei",
+            ref_inv_mass,
+            vec,
+            tagged=(FirstAxisIsElementsTag(),))
```
inducer, line 863: Please use axis tags, not this. Also below.

grudge/op.py
```diff
@@ -819,6 +820,96 @@ def _apply_inverse_mass_operator(
...
+
+    # Compute 1/J * vec
+    def apply_jinv_to_vec(jac_inv, vec):
+        return actx.einsum(
+            "ei,ej->ei",
```
inducer, line 868: Do you mean just `*` here, i.e. element-wise multiplication?  This einsum will sum over each element!

grudge/op.py
```diff
@@ -819,6 +820,96 @@ def _apply_inverse_mass_operator(
...
+        )
+
+    from grudge.geometry import area_element
+
+    if dd_out != dd_in:
+        raise ValueError(
+            "Cannot compute inverse of a mass matrix mapping "
+            "between different element groups; inverse is not "
+            "guaranteed to be well-defined"
+        )
```
inducer, line 837: Why have both `dd_in` and `dd_out`, if having them differ is not appropriate?

grudge/op.py
```diff
@@ -819,6 +820,96 @@ def _apply_inverse_mass_operator(
...
+    ]
+    stage3 = DOFArray(actx, data=tuple(stage3_group_data))
+
+    group_data = [
+        apply_minv_to_vec(
+            reference_inverse_mass_matrix(actx, element_group=grp), vec_i)
+        for grp, vec_i in zip(discr_base.groups, stage3)
+    ]
+
+    return DOFArray(actx, data=tuple(group_data))
```
inducer, line 910: Why build many intermediate `DOFArray`s as opposed to constructing the final one in one go?

grudge/op.py
MTCam: I wondered that myself :).  Copied from the original `_apply_inverse_mass`: https://github.com/inducer/grudge/blob/23fbc05f50500130305aedcd970cdf6609882bb9/grudge/op.py#L798

MTCam: Fixed up in [45d6388](https://github.com/inducer/grudge/pull/354/commits/45d638884251489fca1c732879d9c9d687d7f019)

PR #356, opened 2024-08-04
Fix surface advection example

alexfikl:
The changes to `DOFArray` in https://github.com/inducer/meshmode/pull/421 seem to have slightly broken this example: when doing `dt * rhs`, dt was a `np.array(1.0)` and rhs was `DOFArray` and it wouldn't broadcast that anymore.


inducer:
Thx!

PR #358, opened 2024-08-16
Enable `force_device_scalars` by default

majosm:
Instances of `PyOpenCLArrayContext` that don't explicitly pass `force_device_scalars=True` will have it set to `False`. This is deprecated, and will be an error once inducer/arraycontext#278 goes in.

After the arraycontext PR gets merged I'll submit a follow-up here that makes the interface for `force_device_scalars` match (i.e., allow it to be `None`), so that it will only emit deprecation warnings when explicitly specified.


majosm:
Ruff errors appear to be unrelated.

inducer:
Thx!

PR #361, opened 2024-08-19
Get MPI communicator from array context instead of `dcoll`

majosm:
`dcoll.mpi_communicator` is deprecated.

Note: Can't remove `dcoll.mpi_communicator` yet, because https://github.com/inducer/grudge/pull/285 still needs it.


inducer:
Thx!

PR #369, opened 2024-11-29
Mypy

inducer:
- [ ] Needs https://github.com/inducer/meshmode/pull/443


grudge/array_context.py
```diff
@@ -48,28 +53,6 @@
 
 logger = logging.getLogger(__name__)
 
-try:
-    # FIXME: temporary workaround while SingleGridWorkBalancingPytatoArrayContext
-    # is not available in meshmode's main branch
-    # (it currently needs
-    # https://github.com/kaushikcfd/meshmode/tree/pytato-array-context-transforms)
-    from meshmode.array_context import SingleGridWorkBalancingPytatoArrayContext
```
inducer, line 56: @matthiasdiener This is no longer relevant, right?

matthiasdiener: I think so, yes. We have not been using the `SingleGridWorkBalancingPytatoArrayContext` for quite a while.

PR #375, opened 2025-03-04
Fix some deprecations

alexfikl:
Mostly around `force_device_scalars` and `dataclass_array_container` changes. The remaining warnings are on the pytato side (with too many arguments and unevaluated arguments).


inducer:
Thx!

Issue #223, opened 2022-03-02
Implement proper symbolic tag negotiation for eager


inducer:
- https://github.com/inducer/grudge/pull/222/files#diff-5fa6e6b713d37028c16f1bbc5a6d0b4547ecba0f86c8a52eb30e1ba1dc2614e4R496-R499
- https://github.com/illinois-ceesd/mirgecom/issues/617#issuecomment-1057082716

MTCam:
Just a quick related note... the code here:

https://github.com/inducer/grudge/blob/bde1ad7b0145638b8939152d954bcf3cb4fba73e/grudge/trace_pair.py#L508

fails for array contexts that do not have the `mpi_communicator` attribute, but *do* pass a `comm_tag`.   MIRGE-Com does this when it calls `interior_trace_pairs`, for example here:

https://github.com/illinois-ceesd/mirgecom/blob/452186f0365a5969e55693e07f87caeb712981d6/mirgecom/euler.py#L153

I'll spare you the generated error, but I've just a pushed a drive-by to fix the problem. #226

inducer:
I've started putting some work towards this before realizing that this is all thoroughly pointless: Eager grudge can only have communication for one tag pending at the same time, otherwise it would deadlock. Therefore, all ranks must hit eager communication in the same order, and so we can do with these tags whatever we want; it won't matter until we allow multiple communications to be pending at the same time.

PR #378, opened 2025-05-08
simplify _sym_tag_to_num_tag

matthiasdiener:
- Remove `_TagKeyBuilder` (`KeyBuilder` can handle class types directly now)
- Simplify `base_tag` handling
- Add reference to #223

### DIFF
--- a/grudge/trace_pair.py
+++ b/grudge/trace_pair.py
@@ -89,7 +89,7 @@ from arraycontext import (
 )
 from meshmode.mesh import BTAG_PARTITION, PartID
 from pytools import memoize_on_first_arg
-from pytools.persistent_dict import Hash, KeyBuilder
+from pytools.persistent_dict import KeyBuilder
 
 import grudge.dof_desc as dof_desc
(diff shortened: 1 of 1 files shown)


grudge/trace_pair.py
```diff
@@ -394,33 +389,35 @@ def connected_parts(
...
     from mpi4py import MPI
     tag_ub = MPI.COMM_WORLD.Get_attr(MPI.TAG_UB)
     assert tag_ub is not None
-    key_builder = _TagKeyBuilder()
+    key_builder = KeyBuilder()
     digest = key_builder(comm_tag)
 
     num_tag = sum(ord(ch) << i for i, ch in enumerate(digest)) % tag_ub
 
     warn("Encountered unknown symbolic tag "
```
matthiasdiener, line 414: Perhaps we should downgrade this to logger.info?

inducer: Nope. It's a correctness issue that would be straightforward to avoid.

PR #380, opened 2025-05-23
add MPIEagerJAXArrayContext

matthiasdiener:
~Needs https://github.com/inducer/arraycontext/pull/315 (perhaps)~

_Please squash_


grudge/geometry/metrics.py
```diff
@@ -566,15 +566,21 @@ def _signed_face_ones(
...
         for batch in grp.batches:
             assert batch.to_element_face is not None
             i = actx.to_numpy(actx.thaw(batch.to_element_indices))
-            grp_field = signed_face_ones_numpy[igrp].reshape(-1)
-            grp_field[i] = \
-                (2.0 * (batch.to_element_face % 2) - 1.0) * grp_field[i]
+            sign = (2.0 * (batch.to_element_face % 2) - 1.0)
+            sign_mask[i, :] = sign
+
+        new_group_arrays.append(grp_field * sign_mask)
```
alexfikl, line 580: Why is this multiplying by `grp_field`?

grudge/geometry/metrics.py
```diff
@@ -566,15 +566,21 @@ def _signed_face_ones(
 
     signed_face_ones_numpy = actx.to_numpy(signed_ones)
 
+    new_group_arrays = []
     for igrp, grp in enumerate(all_faces_conn.groups):
+        grp_field = signed_face_ones_numpy[igrp]
+        sign_mask = np.ones_like(grp_field)
```
alexfikl, line 572: I have a hard time remembering why this was written this way (I think those `signed_face_ones` arrays are just created this way so that they have the right shape?), but maybe we can improve it a bit.

How about this?
* Remove the `signed_face_ones` and `signed_face_ones_numpy` things.
* Get the discretization `discr = dcoll.discr_from_dd(dd.with_discr_tag(DISCR_TAG_BASE))`
* For each zip(discr.groups, conn.groups), `mask = np.ones((dgrp.nelements, dgrp.nunit_dofs), dtype=discr.real_dtype)`

Would something like that work? The main idea being that we'll just create everything in numpy and then transfer it over to the array context, so we don't have to worry about how writable the arrays are. 

We might want to tag the end result in the same way that `Discretization.zeros` does too?

alexfikl (commented):
Left a few comments for that `_signed_face_ones` function :grin:

grudge/geometry/metrics.py
matthiasdiener: Yeah, this is redundant for this particular case.

grudge/geometry/metrics.py
matthiasdiener: Thanks for the suggestion! What do you think of 85a0d54? (feel free to push directly to this branch if I did something silly, I am very unfamiliar with this code).

alexfikl: Looks good to me! (doesn't look like the test failure is related?)

PR #382, opened 2025-06-13
Fix Github CI badge links

alexfikl:


alexfikl:
Added these things in #383, so closing.

PR #383, opened 2025-06-13
Remove leap and dagrt dependencies

alexfikl:
This removes the dependency on both `leap` and `dagrt` in `grudge`. 

I added a little compatibility layer so that `set_up_rk4` still works as expected and some other copy paste things from `leap`, but overall not that much.


inducer:
Thanks! Let's proceed with this. I will archive leap and dagrt once this is in.

alexfikl:
> Thanks! Let's proceed with this. I will archive leap and dagrt once this is in.

Sounds good! I'll try to look over why that example is failing tomorrow.

alexfikl:
> > Thanks! Let's proceed with this. I will archive leap and dagrt once this is in.
> 
> Sounds good! I'll try to look over why that example is failing tomorrow.

This was just me deleting things too quickly.. `set_up_rk4` was actually setting up `LSRK4` :grin: It seems to all work now.

PR #384, opened 2025-07-01
Deduplicate in `_dag_to_compiled_func`

majosm:
Needed for inducer/pytato#614.


inducer:
Grudge is in still in quite a rough state, it hasn't made the jump to bpr yet. It's on my list.

PR #388, opened 2025-09-05
add lazy JAX arraycontext

matthiasdiener:
Alternative to #380 / #381

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -34,8 +34,6 @@ intersphinx_mapping = {
 os.environ["PYOPENCL_TEST"] = "port:cpu"
 
 nitpick_ignore_regex = [
-    ["py:mod", r"jax"],  # FIXME: not sure why this does not work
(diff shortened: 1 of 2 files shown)


doc/conf.py
```diff
@@ -33,6 +34,7 @@
 os.environ["PYOPENCL_TEST"] = "port:cpu"
 
 nitpick_ignore_regex = [
+    ["py:mod", r"jax"],  # FIXME: not sure why this does not work
```
alexfikl, line 37: That object doesn't seem to be in their `objects.inv` (from a quick search). Maybe use `jax.numpy` instead? Or just a link `jax <https://docs.jax.dev>`.

alexfikl (approved):
Left a little comment, but otherwise this looks very straightforward now, so not much to say :grin:

inducer:
Thx!

PR #402, opened 2026-03-15
Remove noqa for RET504 (new ruff)

alexfikl:
Can't say I quite understand why this is not an issue anymore. There's still a clear assignment right before (although a long multiline one)..


alexfikl:
Ah, nevermind, apparently both this and TRY003 were just removed from the default set in the new version: https://github.com/astral-sh/ruff/pull/23879.

